PR #31, opened 2019-10-07
return user id on user creation

gt-novelt:
`create_user` returns the newly created user's id.

Id is extracted from the returned request's _Location_ header.


gt-novelt:
@marcospereirampj fyi conflict resolved

marcospereirampj (approved):

PR #54, opened 2020-01-16
Set session `auth` to trivial function to prevent usage of `.netrc` file

joshbode:
Sets the session to not add "basic" auth if entry for keycloak host
exists in users `.netrc` (`requests` causes `Authorization: Basic ...`
header to be added if an entry exists in `.netrc` by default)

see: https://requests.readthedocs.io/en/master/api/#requests.Session.auth

### DIFF
--- a/keycloak/connection.py
+++ b/keycloak/connection.py
@@ -47,6 +47,7 @@ class ConnectionManager(object):
         self._timeout = timeout
         self._verify = verify
         self._s = requests.Session()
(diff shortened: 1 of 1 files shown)


joshbode:
I should have tested this - overriding `auth` doesn't actually work and setting `trust_env` to `False` has the effect of ignoring proxies...

marcospereirampj (approved):

marcospereirampj (approved):

PR #56, opened 2020-02-04
include data in raw_delete request

gitfrosh:
data is missing in raw_delete request: requests with json bodies like in "delete_client_roles_of_user" won't work


pdeutsch:
I was just going to create a pull request for the same thing.   Thanks

Ayodehi:
Likewise, discovered this bug when delete_client_roles_of_user purged all roles! For now, I just let it purge all roles and then restore the desired roles... definitely not ideal.

marcospereirampj (approved):

PR #95, opened 2020-07-05
Fix for KeycloakAdmin refresh_token method

fr33mang:
Hi!

In my work django project we use KeycloakAdmin class and get errors if user session was killed or expired :)

Steps to reproduce:
1) Create KeycloakAdmin class instance
2) Kill user session in Keycloak
3) Try to call methods.
 
It's may be fix!

Recreate session if: 
+ Token is not active
+ Session not active

### DIFF
--- a/keycloak/keycloak_admin.py
+++ b/keycloak/keycloak_admin.py
@@ -1397,7 +1397,13 @@ class KeycloakAdmin:
         try:
             self.token = self.keycloak_openid.refresh_token(refresh_token)
         except KeycloakGetError as e:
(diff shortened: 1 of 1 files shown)


fr33mang:
Can you please leave any comments about the reason of closing?

nicklas-delphai:
I think he closed it because [this pull request](https://github.com/marcospereirampj/python-keycloak/pull/99) was similar to your fix and @marcospereirampj  merged that one instead. Unfortunately that fix is incomplete and doesn't include checking for `Session not active`. I think this change request should be accepted in favor of the other.

PR #194, opened 2021-05-20
Token_only

barbouche:
Allow only token in keycloak_admin object to avoid username and password entry .

### DIFF
--- a/keycloak/__init__.py
+++ b/keycloak/__init__.py
@@ -23,3 +23,4 @@
 
 from .keycloak_admin import *
 from .keycloak_openid import *
+from .keycloak_admin_token import *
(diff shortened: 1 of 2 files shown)


ryshoooo:
Hi! Thanks a lot for your contribution.

To be frank, I'm not quite content with your changes as it introduces an additional class that really does not differ from the admin class in any way, just in the way it initially authenticates. This can be easily handled during class initialization, therefore I'm currently more prone towards #281, as it provides the same functionality within the same class. Could you elaborate a bit on why these changes were made?

ryshoooo:
I'll close this as identical functionality has been merged in from #281

PR #204, opened 2021-06-17
Add function to remove user realm role

bostonk:
here, i have add methods to remove user realm rôle. 

Br


schaarj:
seems to be the same functionality as #201 with a different name and duplication of `URL_ADMIN_USER_REALM_ROLES` as  `URL_ADMIN_DELETE_USER_ROLE`

marcospereirampj (approved):

schaarj:
@marcospereirampj any reason this was merged instead of #201, which seems to me to be somewhat more consistent with the current codebase?

PR #226, opened 2021-09-01
Revert "Fix KeycloakAdmin using wrong realm when authenticating with …

aqiank:
…a service account"

This reverts commit a9b39248543b521fe2a5936fe09b3d8509d681c0.

It seems like it is unnecessary to use master realm to authenticate as admin using service account so this reverts that faulty commit.


iavael:
@marcospereirampj could you please merge this PR or #236?

Issue #93, opened 2020-06-18
I want to add protocol mappers to client 


PreethiNandakumar:
I want to add protocol mappers to client. Any help on this please

pynkpanther:
i managed to do it by adding the mapper via web-admin console first and inspecting the PUT request which get send to the server.  
The payload of the request is exacly the payload you can use for your python-api payload.
When copy / pasting to python code, make sure to replace any `"true"` with the boolean `True` and also sometime you find `true` which you also want to replace with `True`. Same goes for `false`

in my case this was (see in the middle there is the code for the protocolmapper):

```
    keycloak_admin.create_client(
        {
        "id": "6a47220b-8cc2-42d3-be0c-07bb2920962b",
        "clientId": "KEYCLOAK_CLIENT_ID",
        "surrogateAuthRequired": False,
        "enabled": True,
        "alwaysDisplayInConsole": False,
        "clientAuthenticatorType": "client-secret",
        "redirectUris": ["http://localhost"],
        "webOrigins": [],
        "notBefore": 0,
        "bearerOnly": False,
        "consentRequired": False,
        "standardFlowEnabled": True,
        "implicitFlowEnabled": False,
        "directAccessGrantsEnabled": True,
        "serviceAccountsEnabled": False,
        "publicClient": True,
        "frontchannelLogout": False,
        "protocol": "openid-connect",
        "attributes": {
            "saml.assertion.signature": False,
            "saml.force.post.binding": False,
            "saml.multivalued.roles": False,
            "saml.encrypt": False,
            "saml.server.signature": False,
            "saml.server.signature.keyinfo.ext": False,
            "exclude.session.state.from.auth.response": False,
            "saml_force_name_id_format": False,
            "saml.client.signature": False,
            "tls.client.certificate.bound.access.tokens": False,
            "saml.authnstatement": False,
            "display.on.consent.screen": False,
            "saml.onetimeuse.condition": False
        },
        "authenticationFlowBindingOverrides": {},
        "fullScopeAllowed": True,
        "nodeReRegistrationTimeout": -1,
        "protocolMappers": [
            {
            "id": "4ba4e59b-b283-41e2-beca-0870e4ba848d",
            "name": "project_attr_mapper",
            "protocol": "openid-connect",
            "protocolMapper": "oidc-usermodel-attribute-mapper",
            "consentRequired": False,
            "config": {
                "aggregate.attrs": True,
                "multivalued": True,
[...]

PR #232, opened 2021-09-16
Add remove and update methods for client protocol mappers

as3445:
Addresses https://github.com/marcospereirampj/python-keycloak/issues/93


antoniolucasnobar:
Hello @as3445 ,

how can we get the mapper_id to update or delete the mapper?
Is it possible to you add a method to make it easy to get the mapper by its name? Or tell me how to do it.

thank you.

marcospereirampj (approved):

PR #236, opened 2021-10-01
Don't force realm name when using secret key

abbradar:
Using other realms can be useful, for example, to manage realm users with restricted rights.

It also was very unexpected -- I spent some time figuring out why can't I get my access token ;)


CyberzSentry:
Hey. Is there a timeline when is this change going to be pushed to the pip package?

marcospereirampj (approved):

PR #245, opened 2021-11-17
fix invalid credential error and refresh_code not exist error

ziyunxiao:
Hello,

When I use this library to connect to Keycloak V15.0.0 I got two issues.

1. Invalid credentials. 
This error is caused by code in get_token. It chooses the "master" realm if the client_secret_key is provided. While I am not using default "master" real in applications.

2. Refresh token does not exist in token
Somehow when get_token returns it does not include "refresh_token". I am not 100% certain if this is a bug from this library or the something does not work well with the latest version of Keycloak. I tested Keycloak API by postman it does return the "refresh_token". As a workaround, I update the code to call get_token when "refresh_token" does not exist in the token, other than throw an error and quit.

Regards,

Robin


regilero:
I was having the same empty refresh token issue. i found in the current keycloak documentation here https://github.com/keycloak/keycloak-documentation/blob/main/server_admin/topics/clients/oidc/service-accounts.adoc that service accounts does not generate refresh tokens anymore, only access_token. So it may become very important to check this absence of refresh token.

regilero:
https://stackoverflow.com/questions/29233772/why-is-a-refresh-token-not-provided-by-oauth2-servers-responding-to-a-client-cr

This is the reason why no refresh tokens are generated, by default, with service accounts (you could add it in advanced client settings, enabling the 'Use Refresh Tokens For Client Credentials Grant' button).

The Oauth spec states that such B-to-B connection with client_id and secrets should only re-ask for an access_token and as they do not use any final user credentials there is no need for added mechanism of refresh token (as you do not need any user specific credential to get the initial access token, so you can redo it any time).

marcospereirampj (approved):

PR #271, opened 2022-02-28
Fix wrong payload format in README.md samples

bparmentier:
#155 introduced a change in the API in 0.25.0 and payload must now be passed as dict.

Maybe this should've been mentionned in the 0.25.0 changelog as well.


ryshoooo:
Duplicate of https://github.com/marcospereirampj/python-keycloak/pull/375, closing

PR #275, opened 2022-03-05
Token Expiration Issue (see PR description)

FreckTech:
Fix the issue of the token getting expire and not getting refreshed for four functions in the keycloak admin file.


ryshoooo (approved):
LGTM. Thanks for the contribution :)

PR #276, opened 2022-03-05
README.md updates (see PR description) 

FreckTech:
- Added the missed documentation in README.md for keycloak_admin.get_client_authz_settings() function.
- Deleted the documentations for the non existing function keycloak_admin.get_roles() from the README.md.
**_Note:_** it meant to be keycloak_admin.get_realm_roles() instead of keycloak_admin.get_roles() but that is already documented in the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,9 @@ keycloak_admin.get_composite_client_roles_of_user(user_id="user_id", client_id="
 keycloak_admin.delete_client_roles_of_user(client_id="client_id", user_id="user_id", roles={"id": "role-id"})
 keycloak_admin.delete_client_roles_of_user(client_id="client_id", user_id="user_id", roles=[{"id": "role-id_1"}, {"id": "role-id_2"}])
 
+# Get the client authorization settings
+client_authz_settings = get_client_authz_settings(client_id="client_id")
+
 # Get all client authorization resources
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -272,9 +275,6 @@ sync_users(storage_id="storage_di", action="action")
 # Get client role id from name
 role_id = keycloak_admin.get_client_role_id(client_id=client_id, role_name="test")
 
-# Get all roles for the realm or client
-realm_roles = keycloak_admin.get_roles()
```
ryshoooo, line 276: I believe the possibility to get realm roles is still available, it's merely just called `keycloak_admin.get_realm_roles()`. Perhaps we should keep this in the README examples, just fix the method name? :)

FreckTech: @ryshoooo I think you skipped reading my "Note" section in my upper comment 😄

I meant keycloak_admin.get_roles() method doesn't exist, but keycloak_admin.get_realm_roles() exists, and in the README we have both documented which is wrong, so I deleted the documentation for the keycloak_admin.get_roles(), and the other is still there.

The method name seems correct, and I think it should remain keycloak_admin.get_realm_roles(), because we getting the roles for only one realm of a Keycloak server which is defined upon initializing KeycloakAdmin class.

ryshoooo (approved):

PR #297, opened 2022-04-13
UMA-permission requests

Merle-Nerger:
Allow Keycloak to directly evaluate the permissions of a token and decide whether the bearer of a given token should be granted access.

### DIFF
--- a/keycloak/keycloak_openid.py
+++ b/keycloak/keycloak_openid.py
@@ -27,8 +27,8 @@ from jose import jwt
 
 from .authorization import Authorization
 from .connection import ConnectionManager
-from .exceptions import raise_error_from_response, KeycloakGetError, \
-    KeycloakRPTNotFound, KeycloakAuthorizationConfigError, KeycloakInvalidTokenError, KeycloakDeprecationError
+from .exceptions import KeycloakPermissionFormatError, raise_error_from_response, KeycloakGetError, \
+    KeycloakRPTNotFound, KeycloakAuthorizationConfigError, KeycloakInvalidTokenError, KeycloakDeprecationError, KeycloakAuthenticationError
 from .urls_patterns import (
     URL_REALM,
     URL_AUTH,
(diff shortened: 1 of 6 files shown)


keycloak/keycloak_openid.py
```diff
@@ -41,6 +42,24 @@
...
+        self.is_logged_in = is_logged_in
+        self.is_authorized = is_authorized
+        self.missing_permissions = missing_permissions
+
+    def __bool__(self):
+        return self.is_authorized
+
+    def __repr__(self):
+        return f"AuthStatus(is_authorized={self.is_authorized}, is_logged_in={self.is_logged_in}, missing_permissions={self.missing_permissions})"
+
```
ryshoooo, line 62: Might I suggest moving this into the `uma_permissions.py` file?

keycloak/keycloak_openid.py
```diff
@@ -431,3 +455,103 @@ def get_permissions(self, token, method_token_info='introspect', **kwargs):
...
+            if not scopes:
+                needed.discard(resource)
+                continue
+            for scope in scopes:
+                needed.discard("{}#{}".format(resource, scope))
+
+        return AuthStatus(is_logged_in=True, is_authorized=len(needed) == 0, missing_permissions=needed)
+
+
+def build_permission_param(permissions):
```
ryshoooo, line 520: Could this be either moved into `uma_permissions.py` or make it part of the class as a method? I'd just like to avoid having mixing classes and functions in this file :)

ryshoooo, line 520: Could this be either moved into `uma_permissions.py` or make it part of the class as a method? I'd just like to avoid having mixing classes and functions in this file :)

ryshoooo (changes requested):
I'd like to suggest to update with the latest master. The unit test framework was changed into pytest. If you could update your PR with the latest master branch and make sure it passes all checks, then it looks good to me :)

ryshoooo:
Closes as https://github.com/marcospereirampj/python-keycloak/pull/322 resolved the comments

PR #301, opened 2022-05-02
Require minimum urllib 1.26.0.

Wim-De-Clercq:
We use an attribute (allowed_methods) from urllib which is 
only available since 1.26.0 so we should require this as minimum.

Issue: marcospereirampj#196

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     packages=['keycloak', 'keycloak.authorization', 'keycloak.tests'],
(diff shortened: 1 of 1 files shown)


ryshoooo:
Hi! Thanks for your contribution. Your suggestion has been taken to the master branch directly. The only note I have, is that we're actually using `urllib3` instead of `urllib`.

ryshoooo:
Hi! Thanks for your contribution. Your suggestion has been taken to the master branch directly. The only note I have, is that we're actually using `urllib3` instead of `urllib`.

ryshoooo:
Will close this as the requirement has been updated

PR #312, opened 2022-05-19
fix: Add missing keycloak.authorization package

### LINKED ISSUE #311: keycloak.authorization package missing
I think the change in [cc82e6a](https://github.com/marcospereirampj/python-keycloak/commit/cc82e6a8749da1188a8ce57fa2bd9d4e899233b6#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L19) inadvertently removed `keycloak.authorization` from the list of packages, which means it doesn't get included in wheels and doing something like `from keycloak import KeycloakOpenID` fails.

I'll see if I can drop a PR in in a bit.

samdbmg:
Re-adds `keycloak.authorization` to the list of packages in setup.py so it gets included in wheels.

Fixes #311

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
     description="python-keycloak is a Python package providing access to the Keycloak API.",
     long_description=long_description,
     long_description_content_type="text/markdown",
-    packages=["keycloak"],
+    packages=["keycloak", "keycloak.authorization"],
     install_requires=reqs,
     tests_require=dev_reqs,
     extras_require={"docs": docs_reqs},


ryshoooo:
LGTM! Could I ask you to include a simple unit test for this? Importing the KeycloakOpenID or the authorization module should be enough :)

ryshoooo:
> LGTM! Could I ask you to include a simple unit test for this? Importing the KeycloakOpenID or the authorization module should be enough :)

Nah wait sorry it won't. The build process for tox is a bit different. Never mind :) Thanks a lot for fixing this!

ryshoooo (approved):

PR #317, opened 2022-05-20
fix: allow exception to be thrown if role does not exist (i.e. 404) a…

knackjax:
…nd skip exist is true


knackjax:
oh nevermind you fixed already!

PR #322, opened 2022-05-23
Authz

### LINKED ISSUE #318: feat: add request / eval permissions similar to python-keycloak-client
one nice feature I do not see in this library but I see in another keycloak library (python-keycloak-client) is the ability to request / evaluate permissions.

do you want to add this to your library? I can help if you're interested (see https://github.com/Peter-Slump/python-keycloak-client/blob/master/src/keycloak/authz.py#L68-L166)

working on an implementation hopefully available soon

Thanks

knackjax:
updated pull request UMA-permission requests #297 with changes requested by author

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,15 @@ keycloak_openid.load_authorization_config("example-authz-config.json")
 policies = keycloak_openid.get_policies(token['access_token'], method_token_info='decode', key=KEYCLOAK_PUBLIC_KEY)
 permissions = keycloak_openid.get_permissions(token['access_token'], method_token_info='introspect')
 
+# Get UMA-permissions by token
+token = keycloak_openid.token("user", "password")
+permissions = keycloak_openid.uma_permissions(token['access_token'])
+
+# Get auth status for a specific resource and scope by token
+token = keycloak_openid.token("user", "password")
+auth_status = keycloak_openid.has_uma_access(token['access_token'], "Resource#Scope")
+
+
 # KEYCLOAK ADMIN
 
 from keycloak import KeycloakAdmin
--- a/README.md
+++ b/README.md
@@ -114,6 +114,10 @@ permissions = keycloak_openid.get_permissions(token['access_token'], method_toke
 token = keycloak_openid.token("user", "password")
 permissions = keycloak_openid.uma_permissions(token['access_token'])
 
+# Get UMA-permissions by token with specific resource and scope requested
+token = keycloak_openid.token("user", "password")
+permissions = keycloak_openid.uma_permissions(token['access_token'], permissions="Resource#Scope")
+
 # Get auth status for a specific resource and scope by token
 token = keycloak_openid.token("user", "password")
 auth_status = keycloak_openid.has_uma_access(token['access_token'], "Resource#Scope")
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -28,6 +28,7 @@ from jose import jwt
 from .authorization import Authorization
 from .connection import ConnectionManager
 from .exceptions import (
+    KeycloakAuthenticationError,
     KeycloakAuthorizationConfigError,
     KeycloakDeprecationError,
     KeycloakGetError,
@@ -35,6 +36,7 @@ from .exceptions import (
     KeycloakRPTNotFound,
     raise_error_from_response,
 )
+from .uma_permissions import AuthStatus, build_permission_param
 from .urls_patterns import (
     URL_AUTH,
     URL_CERTS,
@@ -46,6 +48,8 @@ from .urls_patterns import (
     URL_USERINFO,
     URL_WELL_KNOWN,
 )
+
+SAME_AS_CLIENT = object()
 
 
 class KeycloakOpenID:
@@ -452,3 +456,75 @@ class KeycloakOpenID:
                     permissions += policy.permissions
 
         return list(set(permissions))
+
+    def uma_permissions(
+        self, token, permissions, audience=SAME_AS_CLIENT, response_mode="permissions"
+    ):
+        """
+        The token endpoint is used to retrieve UMA permissions from Keycloak. It can only be
+        invoked by confidential clients.
+
+        http://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint
+
+
+        :param token: user token
+        :param permissions:
+        :return: permissions list
+        """
+
+        if audience is SAME_AS_CLIENT:
+            audience = self.client_id
+
+        permission = build_permission_param(permissions)
+
+        params_path = {"realm-name": self.realm_name}
+        payload = {
+            "grant_type": "urn:ietf:params:oauth:grant-type:uma-ticket",
+            "permission": permission,
+            "response_mode": response_mode,
+            "audience": audience,
+            "Authorization": "Bearer " + token,
+        }
+
+        self.connection.add_param_headers("Authorization", "Bearer " + token)
+        try:
+            data_raw = self.connection.raw_post(URL_TOKEN.format(**params_path), data=payload)
+        finally:
+            self.connection.del_param_headers("Authorization")
+
+        return raise_error_from_response(data_raw, KeycloakGetError)
+
+    def has_uma_access(self, token, permissions):
+        """
+        Get permission by user token
+
+        :param token: user token
+        :param permissions: dict(resource:scope)
+        :return: result bool
+        """
+        needed = build_permission_param(permissions)
+        try:
+            granted = self.uma_permissions(token, permissions)
+        except (KeycloakGetError, KeycloakAuthenticationError) as e:
+            if e.response_code == 403:
+                return AuthStatus(
+                    is_logged_in=True, is_authorized=False, missing_permissions=needed
+                )
+            elif e.response_code == 401:
+                return AuthStatus(
+                    is_logged_in=False, is_authorized=False, missing_permissions=needed
+                )
+            raise
+
+        for resource_struct in granted:
+            resource = resource_struct["rsname"]
+            scopes = resource_struct.get("scopes", None)
+            if not scopes:
+                needed.discard(resource)
+                continue
+            for scope in scopes:
+                needed.discard("{}#{}".format(resource, scope))
+
+        return AuthStatus(
+            is_logged_in=True, is_authorized=len(needed) == 0, missing_permissions=needed
+        )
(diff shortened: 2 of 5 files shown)


src/keycloak/keycloak_openid.py
```diff
@@ -48,6 +50,9 @@
 )
 
 
+SAME_AS_CLIENT = object()
```
ryshoooo: Could this be moved into `uma_permissions.py` as well? :) It is really used for uma endpoints anyway.

src/keycloak/keycloak_openid.py
```diff
@@ -452,3 +457,63 @@ def get_permissions(self, token, method_token_info="introspect", **kwargs):
...
+
+    def uma_permissions(self, token, permissions, audience=SAME_AS_CLIENT, response_mode="permissions"):
+        """
+        The token endpoint is used to retrieve UMA permissions from Keycloak. It can only be
+        invoked by confidential clients.
+
+        http://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint
+
+
+        :param token: user token
```
ryshoooo, line 468: There are parameters missing and their description can be a bit expanded upon

src/keycloak/keycloak_openid.py
```diff
@@ -452,3 +457,63 @@ def get_permissions(self, token, method_token_info="introspect", **kwargs):
...
+                   "response_mode": response_mode, "audience": audience, "Authorization": "Bearer "+token}
+
+        self.connection.add_param_headers("Authorization", "Bearer " + token)
+        try:
+            data_raw = self.connection.raw_post(
+                URL_TOKEN.format(**params_path), data=payload)
+        finally:
+            self.connection.del_param_headers("Authorization")
+
+        return raise_error_from_response(data_raw, KeycloakGetError)
```
ryshoooo: This should be `KeycloakPostError`

src/keycloak/keycloak_openid.py
```diff
@@ -452,3 +457,63 @@ def get_permissions(self, token, method_token_info="introspect", **kwargs):
                     permissions += policy.permissions
 
         return list(set(permissions))
+
+    def uma_permissions(self, token, permissions, audience=SAME_AS_CLIENT, response_mode="permissions"):
```
ryshoooo: I'm wondering a bit about the `response_mode` parameter. Since this endpoint is supposed to always return permissions, does it make sense to set the response_mode to anything else? If not, we might as well remove it as a method argument.

README.md
```diff
@@ -110,6 +110,15 @@ keycloak_openid.load_authorization_config("example-authz-config.json")
 policies = keycloak_openid.get_policies(token['access_token'], method_token_info='decode', key=KEYCLOAK_PUBLIC_KEY)
 permissions = keycloak_openid.get_permissions(token['access_token'], method_token_info='introspect')
 
+# Get UMA-permissions by token
+token = keycloak_openid.token("user", "password")
+permissions = keycloak_openid.uma_permissions(token['access_token'])
```
ryshoooo, line 115: This is not going to work as the `permissions` argument is required.

src/keycloak/keycloak_openid.py
```diff
@@ -452,3 +457,63 @@ def get_permissions(self, token, method_token_info="introspect", **kwargs):
...
+
+        return raise_error_from_response(data_raw, KeycloakGetError)
+
+    def has_uma_access(self, token, permissions):
+        """
+        Get permission by user token
+
+        :param token: user token
+        :param permissions: dict(resource:scope)
+        :return: result bool
```
ryshoooo: This return value seems to be wrong as we're returning an instance of `AuthStatus`

src/keycloak/keycloak_openid.py
```diff
@@ -452,3 +457,63 @@ def get_permissions(self, token, method_token_info="introspect", **kwargs):
...
+        """
+
+        if audience is SAME_AS_CLIENT:
+            audience = self.client_id
+
+        permission = build_permission_param(permissions)
+
+        params_path = {"realm-name": self.realm_name}
+        payload = {"grant_type": "urn:ietf:params:oauth:grant-type:uma-ticket", "permission": permission,
+                   "response_mode": response_mode, "audience": audience, "Authorization": "Bearer "+token}
```
ryshoooo: I do not believe that the Authorization is needed in the payload based on the Keycloak docs https://www.keycloak.org/docs/latest/authorization_services/index.html#_service_obtaining_permissions

src/keycloak/uma_permissions.py
```diff
@@ -0,0 +1,145 @@
...
+    def __eq__(self, __o: object) -> bool:
+        return str(self) == str(__o)
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    def __hash__(self) -> int:
+        return hash(str(self))
+
+    def __call__(self, *args, resource="", scope="") -> object:
```
ryshoooo: I don't think that the support for multiple args (i.e. the `*` in front of args) is there. We merely just go through all the args and only apply the last one instead of returning a list/tuple of UMAPermissions. So we can either fix this (i.e. support multiple args by returning the same number of permissions) or only allow single arg.

src/keycloak/uma_permissions.py
```diff
@@ -0,0 +1,145 @@
...
+    if isinstance(permissions, (str, UMAPermission)):
+        return set((permissions,))
+
+    try:  # treat as dictionary of permissions
+        result = set()
+        for resource, scopes in permissions.items():
+            if scopes is None:
+                result.add(resource)
+            elif isinstance(scopes, (str, UMAPermission)):
+                result.add("{}#{}".format(resource, scopes))
```
ryshoooo, line 156: In case `scopes` is of type `UMAPermission`, wouldn't this create double `##`? I think this only works if scopes is of type `str`.

src/keycloak/uma_permissions.py
```diff
@@ -0,0 +1,145 @@
...
+            else:
+                for scope in scopes:
```
ryshoooo: This could fail if `scopes` is not iterable. Since we're trying to handle all the exceptions here, we might as well try to handle this one as well :)

src/keycloak/uma_permissions.py
```diff
@@ -0,0 +1,145 @@
...
+                    if not isinstance(scope, (str, UMAPermission)):
+                        raise KeycloakPermissionFormatError(
+                            "misbuilt permission {}".format(permissions)
+                        )
+                    result.add("{}#{}".format(resource, scope))
```
ryshoooo: Same problem here if `scope` is `UMAPermission` type, this will create double `##`.

src/keycloak/uma_permissions.py
```diff
@@ -0,0 +1,145 @@
...
+        return result
+    except AttributeError:
+        pass
+
+    try:  # treat as any other iterable of permissions
+        return set(permissions)
```
ryshoooo: What if permissions is an `Iterable[UMAPermission]`? Then we get a set of instances. I'd suggest `return {str(_perm) for _perm in permissions}`.

ryshoooo (changes requested):
Thanks for taking this up! :)

Overall I think it looks good, I just tried to run possible scenarios in my head and see where things can break. Would be nice to add unit tests for these, but it's not a requirement.

On top of the left comments, if you could go through the docstrings as well and make sure they are updated and present. I could see the UMAPermission classes are missing quite a bit of them.

Thanks a lot for the contribution! :) Feel free to discuss/oppose the opinions in the comments as I might just as well be wrong in them :)

ryshoooo:
FYI If you're having trouble with the formatting checks, I can refer you to the CONTRIBUTING guidelines https://github.com/marcospereirampj/python-keycloak/blob/master/CONTRIBUTING.md#running-checks-and-tests.

src/keycloak/keycloak_openid.py
knackjax: sounds good to me, moving to `uma_permissions.py`

knackjax: at a closer look, I can eliminate this parameter, I do not see why it would be anything besides client_id.

src/keycloak/keycloak_openid.py
knackjax: fixed.

src/keycloak/keycloak_openid.py
knackjax: agreed, this parameter has been removed.

README.md
knackjax: I have made permissions optional and added more examples

src/keycloak/keycloak_openid.py
knackjax: correct, I fixed this

src/keycloak/keycloak_openid.py
knackjax: agreed, removed.

src/keycloak/keycloak_openid.py
knackjax: added more description

src/keycloak/uma_permissions.py
knackjax: makes sense, I made it only accept a single arg now

src/keycloak/uma_permissions.py
knackjax: correct, only accept str now

src/keycloak/uma_permissions.py
knackjax: added case to handle non-iterables

src/keycloak/uma_permissions.py
knackjax: fixed

src/keycloak/uma_permissions.py
knackjax: good catch, i do an iterable check and added more unit tests

knackjax:
> Thanks for taking this up! :)
> 
> Overall I think it looks good, I just tried to run possible scenarios in my head and see where things can break. Would be nice to add unit tests for these, but it's not a requirement.

Apologies, I forgot to add the unit test to the commit, now it is included.

> On top of the left comments, if you could go through the docstrings as well and make sure they are updated and present. I could see the UMAPermission classes are missing quite a bit of them.

 Ok I added some missing docstrings

> Thanks a lot for the contribution! :) Feel free to discuss/oppose the opinions in the comments as I might just as well be wrong in them :)

No problem, thanks I like all your suggestions.

knackjax:
> FYI If you're having trouble with the formatting checks, I can refer you to the CONTRIBUTING guidelines https://github.com/marcospereirampj/python-keycloak/blob/master/CONTRIBUTING.md#running-checks-and-tests.

thanks! my computer is locked down so I couldn't run the unit tests (i.e. with docker), `tox -e check` was useful.

ryshoooo (approved):

PR #281, opened 2022-03-17
Fixes client credentials grant for KeycloakAdmin

andricDu:
# Summary

This is a fix for the following issue that describes an bug when using client credentials with the `KeycloakAdmin` client: https://github.com/marcospereirampj/python-keycloak/issues/269

The `if` conditions are restructured in the `get_token` method such that client_credentials and password grant types are both exclusive and properly executed. Common token and header logic between grant types is encapsulated by the inner function. 

## Validation
Verified with the following script and running latest keycloak via docker:
```
docker run -p 8080:8080 -e KEYCLOAK_ADMIN=admin -e KEYCLOAK_ADMIN_PASSWORD=admin quay.io/keycloak/keycloak:17.0.0 start-de
```

### Script
```python
from keycloak import KeycloakAdmin

print("\n****** ADMIN USER ******")
admin = KeycloakAdmin(server_url="http://localhost:8080", username="admin", password="admin")
print(admin.token)

print("\n\n****** SERVICE ACCOUNT ******")
service_account = KeycloakAdmin(server_url="http://localhost:8080", client_id="test", client_secret_key="2pZ8gI2jhSGMHi0RPrRle1qOIteRAsrE")
print(service_account.token)
```

### Output
```
$ python verify.py 

****** ADMIN USER ******
[...]

Issue #269, opened 2022-02-17
Regression error on 0.27 for client_credentials flow


AimerLAir:
Hi, it seems that the **client_credentials flow** doesn't work anymore.
The **get_token** method doesn't request token if there is no username and password, what i believe it should do anyway if grant_type is **["client_credentials"]**.

```
def get_token(self):
        token_realm_name = 'master' if self.client_secret_key else self.user_realm_name or self.realm_name
        self.keycloak_openid = KeycloakOpenID(server_url=self.server_url, client_id=self.client_id,
                                              realm_name=token_realm_name, verify=self.verify,
                                              client_secret_key=self.client_secret_key,
                                              custom_headers=self.custom_headers)

        grant_type = ["password"]
        if self.client_secret_key:
            grant_type = ["client_credentials"]
            if self.user_realm_name:
                self.realm_name = self.user_realm_name

        if self.username and self.password:
            self._token = self.keycloak_openid.token(self.username, self.password, grant_type=grant_type)

            headers = {
                'Authorization': 'Bearer ' + self.token.get('access_token'),
                'Content-Type': 'application/json'
            }
        else:
            self._token = None
            headers = {}

        if self.custom_headers is not None:
            # merge custom headers to main headers
            headers.update(self.custom_headers)

        self._connection = ConnectionManager(base_url=self.server_url,
                                             headers=headers,
                                             timeout=60,
                                             verify=self.verify)
```

This behavior was introduced by the #244. (But i do not understand what was its purpose.)

j-hellenberg:
We've also noticed this issue. As you said, currently no token is ever requested for the `client_credentials` flow, which means the library is practically unusable for this authentication method (except by passing some dummy username and password, which will be ignored due to `grant_type` == `["client_credentials"]` but still trigger the if condition...).

Judging by its description, I believe what #244 was trying to achieve was to allow using the library without having to specify any credentials (`username/password` or `client_id/client_secret_key`) at all by directly authenticating using a valid authentication token passed into the `custom_headers` field of the `KeycloakAdmin` constructor (previously, the `get_token` method would throw an error if not provided with any credentials at all, as the call of `self.keycloak_openid.token` would fail in that case). I'm not quite sure why somebody would want to do that though, as the manually provided auth token would eventually expire, at which point the library hasn't got the credential data required to initiate a re-authentication.

[...]

PR #324, opened 2022-05-24
fix: allow client_credentials token if username and password not spec…

knackjax:
…ified

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -2566,13 +2566,15 @@ class KeycloakAdmin:
             custom_headers=self.custom_headers,
         )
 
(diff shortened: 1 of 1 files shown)


knackjax:
I noticed I could not use client_id + secret_key with grant client_credentials to get a token back for keycloak admin. This will fix it. An example how this works here https://www.appsdeveloperblog.com/keycloak-client-credentials-grant-example/. Please let me know if you have questions.

knackjax:
looks like its related to issue https://github.com/marcospereirampj/python-keycloak/issues/269 and PR https://github.com/marcospereirampj/python-keycloak/pull/281

ryshoooo (approved):
LGTM :)

Thanks for fixing this!

PR #334, opened 2022-06-02
test: Make curl silent

asyd:
Prevent curl complaining while keycloak is starting, and remove HTML dump


ryshoooo (commented):
LGTM, I'd just suggest instead of `feat:` for the commit for this to be `chore:` or `test:`, no need to release a new version when only test framework changes are made :)

ryshoooo (approved):

PR #346, opened 2022-06-22
feat: Ability to set custom timeout for KeycloakOpenId and KeycloakAdmin

fredlb:
This adds the ability to set a custom timeout for all requests sent by KeycloakOpenID and KeycloakAdmin.
Since KeycloakAdmin constructs an instance of KeycloakOpenID, its hard to modify that timeout (you still can by doing `keycloak_admin.keycloak_openid._connection._timeout = ...`)


ryshoooo (approved):
Thanks a lot for your contribution :)

All looks good!

PR #350, opened 2022-06-27
feat: use poetry for package management

### LINKED ISSUE #349: Conflicting dependency management
There are multiple tools for dependency and package management setup. Pipenv, setup tools, pep517/8, requirements file.

These are confusing. Can there be just one in place? A single source of truth.

Personally, I prefer PEP517 and I use poetry for it. It covers everything including locking 🔑.

chumaumenze:
There are multiple tools for dependency and package management setup. Pipenv, setup tools, pep517/8, requirements file.

These are confusing. Can there be just one in place? A single source of truth.

Personally, I prefer PEP517 and I use poetry for it. It covers everything including locking 🔑.

This PR introduces poetry for package management. Fixes #349

Tests and tox are working as before.


CONTRIBUTING.md
```diff
@@ -10,11 +10,17 @@ The development environment is mainly up to the developer. Our recommendations a
...
-python -m pip install -r requirements.txt
-python -m pip install -r dev-requirements.txt
+# Install and upgrade pip & poetry
+python -m pip install --upgrade pip poetry
+
+# Create virtualenv
+python -m poetry env use <PATH_TO_PYTHON_VERSION>
+
+# install package dependencies including dev dependencies
+python -m poetry install --dev
```
ryshoooo: This does not seem to be right, there is no `--dev` option (https://python-poetry.org/docs/cli/#install)

pyproject.toml
```diff
@@ -1,3 +1,74 @@
+[tool.poetry]
+name = "python-keycloak"
+version = "1.9.0"
```
ryshoooo: Can we keep this as `0.0.0`? Just that it's clear it's a "dummy" label in here, it gets overwritten in the publish pipeline anyway.

ryshoooo (changes requested):
Hi @chumaumenze 

Thanks a lot for your contribution! The changes look good and are pretty simple to follow :)
I've left 2 small comments. Also, the pipeline seems to be failing for Python versions 3.8 and 3.9, if you could take a look at that as well :)

I'd also ask if you could change the commit message from `feat: ...` to `chore: ...` or `build: ...` as this is more of a new DevOps feature rather than a new library feature. Otherwise great work, works really smoothly :)

CONTRIBUTING.md
chumaumenze: Nice catch. Thank you

ryshoooo (approved):

PR #353, opened 2022-07-12
fix: Support the auth_url method called with scope & state params now

### LINKED ISSUE #206: where is the api endpoint for authentication
where can i find the function for open authentication, I tried `auth_url`, the server logs:
```
17:06:58,084 WARN  [org.keycloak.services] (default task-2) KC-SERVICES0091: Request is missing scope 'openid' so it's not treated as OIDC, but just pure OAuth2 request.

``` 

the server recognize this request as OAuth2 since it doesn't specify the `scope` parameter, should I custimze this method to include `scope=id` to make the authentication work?

Zerek-Cheng:
We need the standard Oauth 2.0 Protocols 's scope & state params to allow the most import function!
the oauth 2.0 docs: 
> https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint
```
scope
REQUIRED. OpenID Connect requests MUST contain the openid scope value. If the openid scope value is not present, the behavior is entirely unspecified. Other scope values MAY be present. Scope values used that are not understood by an implementation SHOULD be ignored. See Sections [5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims) and [11](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess) for additional scope values defined by this specification.
state
RECOMMENDED. Opaque value used to maintain state between the request and the callback. Typically, Cross-Site Request Forgery (CSRF, XSRF) mitigation is done by cryptographically binding the value of this parameter with a browser cookie.
```
that's important

### DIFF
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -176,10 +176,16 @@ class KeycloakOpenID:
 
     def auth_url(self, redirect_uri, scope="email", state=""):
         """
+        Get authorization URL endpoint.
 
-        http://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint
-
-        :return:
+        :param redirect_uri: Redirect url to receive oauth code
+        :type redirect_uri: str
+        :param scope: Scope of authorization request, split with the blank space
+        :type: scope: str
+        :param state: State will be returned to the redirect_uri
+        :type: str
+        :returns: Authorization URL Full Build
+        :rtype: str
         """
         params_path = {
             "authorization-endpoint": self.well_known()["authorization_endpoint"],
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -174,7 +174,7 @@ class KeycloakOpenID:
 
         return raise_error_from_response(data_raw, KeycloakGetError)
 
-    def auth_url(self, redirect_uri):
+    def auth_url(self, redirect_uri, scope="email", state=""):
         """
 
         http://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint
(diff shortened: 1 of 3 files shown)


src/keycloak/keycloak_openid.py
```diff
@@ -174,7 +174,7 @@ def well_known(self):
 
         return raise_error_from_response(data_raw, KeycloakGetError)
 
-    def auth_url(self, redirect_uri):
+    def auth_url(self, redirect_uri, scope="email", state=""):
```
ryshoooo, line 177: Could I ask you to add these into the docstring? :)

ryshoooo (commented):
Overall looks good :) Just a single comment left. Also if I can ask you to rename the `feat:` commit into `fix:` instead, no need to bump the major version for this change.

Zerek-Cheng:
@ryshoooo  @marcospereirampj  
commit record rebase done
thks,this is my first time to PR
i am learning

and...why didn't I see auth_url in the old document

src/keycloak/keycloak_openid.py
Zerek-Cheng: > Could I ask you to add these into the docstring? :)

done!

ryshoooo: By docstring I meant the method's docstring, not the README file :) i.e. something like

```python
def auth_url(self, redirect_uri, scope="email", state=""):
    """
    Get authorization URL endpoint.

    :param redirect_uri: Redirect url
    :type redirect_uri: str
    :param scope: Scope
    :type: scope: str
    :param state: State
    :type: str
    :returns: Authorization URL
    :rtype: str
    """"
    ...
```

Zerek-Cheng: Thank you for reminding me.
But I'm in bed now.I will deal with it after I get up.

Zerek-Cheng: @ryshoooo 
 Everything Done！
 Check pls！

ryshoooo (approved):
LGTM :) Thanks a lot for your contribution!

PR #371, opened 2022-08-19
feat: add client scope-mappings client roles operations

antoniolucasnobar:
Add client-level roles to the client’s scope

POST /{realm}/client-scopes/{id}/scope-mappings/clients/{client}

Get the roles associated with a client’s scope Returns roles for the client.

GET /{realm}/client-scopes/{id}/scope-mappings/clients/{client}

Remove client-level roles from the client’s scope.

DELETE /{realm}/client-scopes/{id}/scope-mappings/clients/{client}

Please let me know if you need anything to be changed.


ryshoooo (approved):
LGTM :) Thanks a lot!

PR #374, opened 2022-09-02
Docs: Change to username

istiak101:
I was trying to get an user with the email address but this actually fetches data using the username. A user can be both email address as username and username string depending on realm configuration. My realm config was username string. So rewriting this to remove confusion.


ryshoooo (approved):
Hi @istiak101 

Thanks a lot, this change is correct :) All looks good, just make sure your commit message follows conventional commits (preferably starts with `docs:`).

istiak101:
Thanks @ryshoooo 

I have updated the commit message.

PR #397, opened 2022-12-17
Feat disable user

iglimanaj:
This pull request introduces a few helper functions for dealing with users in Keycloak. The are handy if you need to enable or  disable users programmatically.

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -762,6 +762,28 @@ class KeycloakAdmin:
         )
         return raise_error_from_response(data_raw, KeycloakPutError, expected_codes=[204])
 
+    def disable_user(self, user_id):
+        """Disable the user from the realm. Disabled users can not log in.
+
+        :param user_id: User id
+        :type user_id: str
+
+        :return: Http response
+        :rtype: bytes
+        """
+        return self.update_user(user_id=user_id, payload={"enabled": False})
+
+    
+    def disable_all_users(self):
+        """Disable all existing users.
+        """
+        users = self.get_users()
+        for user in users:
+            user_id = user["id"]
+            print(f"Disabling user with id: {user_id}")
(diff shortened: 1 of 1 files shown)


src/keycloak/keycloak_admin.py
```diff
@@ -762,6 +762,46 @@ def update_user(self, user_id, payload):
...
+        """
+        return self.update_user(user_id=user_id, payload={"enabled": True})
+    
+    def disable_all_users(self):
+        """Disable all existing users.
+        """
+        users = self.get_users()
+        for user in users:
+            user_id = user["id"]
+            print(f"Disabling user with id: {user_id}")
```
ryshoooo: No `print`s please :) I do not want to force unstoppable logs to the users of the library. You can use `logging` with a combination of a `verbose` parameter or just a `verbose` parameter itself keeping the prints, but the default should be 0 verbosity :)

src/keycloak/keycloak_admin.py
```diff
@@ -762,6 +762,46 @@ def update_user(self, user_id, payload):
...
+            self.disable_user(user_id=user_id)
+        
+    def enable_all_users(self):
+        """Disable all existing users.
+        """
+        users = self.get_users()
+        for user in users:
+            user_id = user["id"]
+            print(f"Enabling user with id: {user_id}")
```

ryshoooo (commented):
Hi @iglimanaj 

Thanks a lot for your contribution :) LGTM just a small comment on the `print` statements.

src/keycloak/keycloak_admin.py
iglimanaj: Good point. I removed the print statements.

ryshoooo:
Looks like the pipeline is failing due to incorrectly formatted code. Make sure that `tox -e check` passes.

iglimanaj:
@ryshoooo I just pushed the refactored code. Thanks for your review!

PR #400, opened 2022-12-28
Feat/create permission for scopes

Hadeer-Elsaeed:


src/keycloak/keycloak_admin.py
```diff
@@ -3460,6 +3460,36 @@ def get_client_authz_scope_permission(self, client_id, scope_id):
         )
         return raise_error_from_response(data_raw, KeycloakGetError)
 
+    def create_client_authz_scope_permission(self, payload, client_id):
```
ryshoooo, line 3862: Hi @Hadeer-Elsaeed 

Would you mind adding a unit test for this one? I just want to make sure that it works as intended with the latest Keycloak version :)

Hadeer-Elsaeed: Hi @ryshoooo 
I have already added unit test. could you check this
https://github.com/marcospereirampj/python-keycloak/pull/400/commits/99089c190b2088a3fd599a0ef168c880b952d374

ryshoooo: Looks good :) can you push it to this branch?

Hadeer-Elsaeed: @ryshoooo  I push it but I think there is other problem, the pipeline failed and also the pipeline ""Daily Check" failed yesterday for the same reason. could you check it ?

Hadeer-Elsaeed: sorry @ryshoooo to make a lot of failed actions, I have a  problem to run tests local if you can help .. when I run command "tox -e tests" .. Its created the container and gets "Timeout exceeded" without running testcases ?

ryshoooo: So do you have `docker` running? If so, ensure you don't have anything running at port 8080. Otherwise, just post all of the output you are getting, there should be some indication of what is going wrong :)

ryshoooo (approved):

ryshoooo:
Looks like the newly included test is failing at

```
permission_name = admin.get_client_authz_scope_permission(
>           client_id=realm_management_id, scope_id=token_exchange_scope_id
        )["name"]
```

where Keycloak responds with a 404.

PR #403, opened 2023-01-24
feat: add Keycloak UMA client

nuwang:
Add support for a UMA client for keycloak.
Currently, support for creating resources is included, along with associated tests.


ryshoooo (approved):
Very nicely done @nuwang! Thanks a lot :)

PR #406, opened 2023-02-02
feat: init KeycloakAdmin with token

### LINKED ISSUE #404: Add option to authenticate admin via access token
Currently the way to authenticate an admin is by providing `username` and `password`. I suggest adding an option to authenticate the user using an access token. Something like the following:
```python
from keycloak import KeycloakAdmin

keycloak_admin = KeycloakAdmin(server_url="http://localhost:8080/auth/",
                               access_token='admin-access-token',
                               realm_name="master",
                               user_realm_name="only_if_other_realm_than_master",
                               client_secret_key="client-secret",
                               verify=True)

```

It seems that `KeycloakAdmin.__init__()` is fetching an access token. I think we can skip that step if we already got the token.
There are even scenarios where an admin user would not be able to be authenticated currently, for example, if the only way to login was via an external IDP. In those cases we only got the access token.

I can work on this feature if it gets a green flag.

JeronimoMendes:
closes #404

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -157,10 +157,14 @@ class KeycloakAdmin:
         if self.token is None:
             self.get_token()
 
-        headers = {
-            "Authorization": "Bearer " + self.token.get("access_token"),
-            "Content-Type": "application/json",
-        } if self.token is not None else {}
+        headers = (
+            {
+                "Authorization": "Bearer " + self.token.get("access_token"),
+                "Content-Type": "application/json",
+            }
+            if self.token is not None
+            else {}
+        )
 
         if self.custom_headers is not None:
             # merge custom headers to main headers
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -154,9 +154,21 @@ class KeycloakAdmin:
         self.custom_headers = custom_headers
         self.timeout = timeout
 
-        # Get token Admin
-        if not self.token:
+        if self.token is None:
             self.get_token()
+
+        headers = {
+            "Authorization": "Bearer " + self.token.get("access_token"),
+            "Content-Type": "application/json",
+        } if self.token is not None else {}
+
+        if self.custom_headers is not None:
+            # merge custom headers to main headers
+            headers.update(self.custom_headers)
+
+        self.connection = ConnectionManager(
+            base_url=self.server_url, headers=headers, timeout=60, verify=self.verify
+        )
 
     @property
     def server_url(self):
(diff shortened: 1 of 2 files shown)


JeronimoMendes:
@jreusch does this PR solve your problem as well?

jreusch:
> @jreusch does this PR solve your problem as well?

thanks for the PR, you were faster, i still did not manage to get to the tests.

i'm just wondering? how does this work?
this changes together when a token is passed leads to `self.get_token()` not beeing  called, but that's the only place were i see `self.connection` beeing initialized (and the headers set)?

JeronimoMendes:
I forgot to commit the header update 😅. I'm currently just doing in inside the `__init__` but I think there's a better way of encapsulating this logic as it is also used on `get_token()`. I'll be pushing the changes later.

I'm still not sure how to test this, they currently check the init by checking if there's a token on the object. In this case, the token is already provided 🫤

jreusch:
hey hey

> I forgot to commit the header update sweat_smile. I'm currently just doing in inside the `__init__` but I think there's a better way of encapsulating this logic as it is also used on `get_token()`. I'll be pushing the changes later.

thanks for adapting! i'll have a look at them now, coming back to you.

> I'm still not sure how to test this, they currently check the init by checking if there's a token on the object. In this case, the token is already provided 🫤

as said, i've already digged into the tests, i had problems getting them to run here and could not continue since them, currently tied up in moving flats :/

my idea for testing was: all the happy init flows seem to be combined into *one* testcase where each of them is tested after another.

yes those seem not to check e.g. if a connection is existing and only what you're describing, that's why i'd maybe as first go with "take the token which is asserted in a previous test, make a new init with that token and check if the token is set.

plus could be to finally also check if the connection was set up properly, then this point on your first change would have been spotted immediatelly :)

ryshoooo:
Thanks a lot @JeronimoMendes! :) LGTM!

yalattas:
nice, this is exactly where I am stuck. I am creating a user from the backend client using keycloak admin credentials. It works for a while, then after around 30 minutes. The admin session started to be expired therefore, I need to re-initiate the client again. Otherwise, keycloak is responding with Unauthorized something like that.

Thanks for solving it. But quick question though, how can I pass a token now instead of username/password and from where can I find the token?
Apologies, I am a bit new to IAM implementation

ryshoooo:
> nice, this is exactly where I am stuck. I am creating a user from the backend client using keycloak admin credentials. It works for a while, then after around 30 minutes. The admin session started to be expired therefore, I need to re-initiate the client again. Otherwise, keycloak is responding with Unauthorized something like that.
> 
> Thanks for solving it. But quick question though, how can I pass a token now instead of username/password and from where can I find the token? Apologies, I am a bit new to IAM implementation

Hi @yalattas 

What you're searching for is to use the KeycloakAdmin class with the auth_refresh_token parameter set (https://github.com/marcospereirampj/python-keycloak/blob/master/src/keycloak/keycloak_admin.py#L109), so something like
```
admin = KeycloakAdmin(server_url="https://my.keycloak.host/auth", username="admin_username", password="admin_password", realm_name="realm", auto_refresh_token={"get", "post", "put", "delete"})
```
Then every time on any of the GET, POST, PUT, or DELETE requests if the token is not active, the client will automatically refresh it for you.

Issue #380, opened 2022-10-19
Client Secret Key of Exchange Token Reques


ioshumer:
In key of using 'openid-connect/token' resource of a realm Keycloak requires client_id/client_secret combination in a request's body. Here, in the method `echange_token` of the lib following code:

> payload = self._add_secret_key(payload)

Method `self._add_secret_key` adds  *client_secret* of a KeycloakOpenID object which is using as a **starting client** for token exchange instead of adding **target's client**  *secret*.

Would it be better to append **client_secret** parameter to `exchange_token` method?

ryshoooo:
Hi @ioshumer. Well, I'm not sure, to be honest, I haven't quite played around with confidential clients and their exchanges. I'd expect that you need to pass on both secrets, i.e. for both clients when you wish to make an exchange from one to another, which would then require an additional argument for the exchange token method. But I might be mistaken :)

It needs some experimentation from my side, but otherwise sounds like a reasonable request :)

remicres:
I agree with @ioshumer `client_secret: str` is missing in the `exchange_token()` args.

For instance if the goal is to exchange the source_client token with a token for the target_client that can be refreshed by this last, the exchange has to be initiated from the target_client. In this case must be provided during the exchange:
- the target client_id
- the target client_secret
- the source client access token 

e.g.
```py
data = {
    "client_id": tgt_client_id,
    "client_secret": tgt_client_secret,
    "grant_type": "urn:ietf:params:oauth:grant-type:token-exchange",
    "subject_token": src_access_token,
    "requested_token_type": "urn:ietf:params:oauth:token-type:refresh_token",
}
exchanged = requests.post(token_endpoint, data=data).json()
```

PR #411, opened 2023-02-20
Feature/exchange token

AndreasBBS:
# Suggesting improvements to exchange token feature

I've been using the exchange token feature of Keycloak. While trying to use this library I ran into an issue, I couldn't pass the requested_issuer field. This is a requirement when exchanging the access token for an IDP token. For this reason I decided to make some changes to the implementation which I'm suggesting to merge through this pull request.

## Suggested changes

- Add missing arguments, subject_token_type, subject_issuer and requested_issuer. These exist in the [specification](https://www.keycloak.org/docs/19.0.3/securing_apps/#_token-exchange) but were not possible to set. The requested_issuer is especially important to allow exchanging for an IDP token

- Make some of the optional fields type's Optional and give them None for the default value

- Use the class's inner client_id instead of receiving it from the arguments. If the library user needs a token from a specific client (not the subject client) he should initialize an instance with that client

## Notes

It's not big changes but in my opinion it greatly improves the functionality of this feature in a way that takes full advantage of keycloak's full implementation. Anybody should be able to now exchange their access tokens for the IDP's tokens for example if you want Google's token you just need to do:
```
keycloak_openid.exchange_token(
    token=access_token,
    requested_token_type='urn:ietf:params:oauth:token-type:access_token',
    requested_issuer='google'
)
```

Let me know if you like the idea and if it's something you'd like to add to this package. I'll be eagerly awaiting all the constructive criticism of my contributions!

### DIFF
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -28,6 +28,7 @@ class to handle authentication and token manipulation.
 """
 
 import json
+from typing import Optional
 
 from jose import jwt
 
@@ -341,9 +342,11 @@ class KeycloakOpenID:
     def exchange_token(
         self,
         token: str,
-        client_id: str,
         audience: str,
-        subject: str,
+        subject: Optional[str] = None,
+        subject_token_type: Optional[str] = None,
+        subject_issuer: Optional[str] = None,
+        requested_issuer: Optional[str] = None,
         requested_token_type: str = "urn:ietf:params:oauth:token-type:refresh_token",
         scope: str = "openid",
     ) -> dict:
@@ -354,12 +357,16 @@ class KeycloakOpenID:
 
         :param token: Access token
         :type token: str
-        :param client_id: Client id
-        :type client_id: str
         :param audience: Audience
         :type audience: str
         :param subject: Subject
         :type subject: str
+        :param subject_token_type: Token Type specification
+        :type subject_token_type: Optional[str]
+        :param subject_issuer: Issuer
+        :type subject_issuer: Optional[str]
+        :param requested_issuer: Issuer
+        :type requested_issuer: Optional[str]
         :param requested_token_type: Token type specification
         :type requested_token_type: str
         :param scope: Scope, defaults to openid
@@ -370,11 +377,14 @@ class KeycloakOpenID:
         params_path = {"realm-name": self.realm_name}
         payload = {
             "grant_type": ["urn:ietf:params:oauth:grant-type:token-exchange"],
-            "client_id": client_id,
+            "client_id": self.client_id,
             "subject_token": token,
+            "subject_token_type": subject_token_type,
+            "subject_issuer": subject_issuer,
             "requested_token_type": requested_token_type,
             "audience": audience,
             "requested_subject": subject,
+            "requested_issuer": requested_issuer,
             "scope": scope,
         }
         payload = self._add_secret_key(payload)
--- a/tests/test_keycloak_openid.py
+++ b/tests/test_keycloak_openid.py
@@ -210,7 +210,6 @@ def test_exchange_token(
     # Exchange token with the new user
     new_token = oid.exchange_token(
         token=token["access_token"],
-        client_id=oid.client_id,
         audience=oid.client_id,
         subject=username,
     )


src/keycloak/keycloak_openid.py
```diff
@@ -370,11 +378,14 @@ def exchange_token(
         params_path = {"realm-name": self.realm_name}
         payload = {
             "grant_type": ["urn:ietf:params:oauth:grant-type:token-exchange"],
-            "client_id": client_id,
+            "client_id": self.client_id,
```
ryshoooo, line 380: I´m not sure I´m entirely on board with this change :) Removing `client_id` from the input arguments is a breaking change, we will be forcing users to rewrite their code with this change (including myself as I'm heavily dependent on this method). Which by itself is not a bad thing, but I want to have a clear justification for this change.

The reason why I'm not fully on board, is that the client id is tightly linked to the token, i.e. the token has to be generated by the same client. So if you pass a token not generated by the client, you don't really have a simple way to change that. On the other hand, I do understand it from the encapsulation concept of the KeycloakOIDClient class. But we are reducing the capabilities of this method now, it is not "just exchange any token to any token", now it is becoming "exchange current client's tokens only to other tokens".

AndreasBBS: I completely understand and empathize with not wanting to introduce a breaking change. I can remove this change and we can discuss introducing it in a different pull-request.

This being said I still believe it's an important change in terms of the encapsulation of the KeycloakOIDClient class. If you want to have a method that allows to "just exchange any token to any token" I believe it should be a static method and the client_secret needs to be an argument or you'll run into problems like this the one in this issue  #380. In my opinion this method should be a "exchange current client's tokens only to other tokens" type of behavior, especially because in its implementation it gets the client secret from the instance. (`payload = self._add_secret_key(payload)`)

For this reasons I believe it needs to be either a static method of the class, and both the client_id and client_secret need to be arguments of the static method, or an instance method that gets the client_id from the instance.

Either way, as I mentioned in the beginning of this reply, this is not the main subject of this pull request. My main goal is to be able to use the requested_issuer field to exchange my token for an IDP token. And so, if you want to move this to a different Pull Request I'm more than happy to make the necessary changes and submit this specific alteration in a different Pull Request.

ryshoooo: Nah I think you're right :) It makes more sense the way you propose. I think it's fine, let's just make sure the conventional commit message contains the BREAKING CHANGE indication :)

AndreasBBS: Sounds good. I'm a bit busy but when I have a bit of free time I'll do that.
I'll need to change some of the unit tests that are failing relating to this method. Should I include that message in that particular commit or do I have to put in both?

ryshoooo: Only 1 is enough :)

AndreasBBS: Sorry for the long time it took me to get back to this. I've been busy as hell but this was not forgotten! Happy to be able to contribute and I hope my contribution is well appreciated :smile:

ryshoooo (approved):
Hi @AndreasBBS 

Thanks a lot for this, LGTM :)
Unfortunately, however, I can't merge it just yet. Merging this will release python-keycloak 3.0.0 and we're claiming that python-keycloak 3.0.0 removes the deprecated attributes in the KeycloakAdmin object. So either we include the removal of the deprecated attributes in this PR or we update the deprecated message stating that these attributes will be removed in 4.0.0 or we merge it later after the deprecated attributes are released.

AndreasBBS:
I don't know if moving from 3.0 to 4.0 too fast is ideal. You can include the removal of the deprecated attributes in this PR or include this PR in the branch with the removal of the deprecated attributes and squash it if needed. Right now I'm using my own branch in a project where I need this change but it would be nice to be able to use the main package. How can I help?

ryshoooo:
> I don't know if moving from 3.0 to 4.0 too fast is ideal. You can include the removal of the deprecated attributes in this PR or include this PR in the branch with the removal of the deprecated attributes and squash it if needed. Right now I'm using my own branch in a project where I need this change but it would be nice to be able to use the main package. How can I help?

Perhaps I'll postpone the deprecation of the attributes into 4.0.0 :) Also giving more time for the change. I'll prepare the change for that, when ready, I'll merge this in :)

ryshoooo (approved):
Thanks a lot @AndreasBBS for your contribution! :)

PR #415, opened 2023-02-28
refactor: Refactor auto refresh

nuwang:
This PR factors out a KeycloakOpenIDConnectionManager class from the keycloak_admin client, which is responsible for automatically refreshing tokens as required. This class is now used by both the keycloak_admin client and the keycloak_uma client, so that both clients are able to auto refresh tokens as they expire, and it is no longer necessary to decide when to refresh on a per-method basis. It also enables better separation of concerns.

#### features and backward compatibility
* The PR does not break backward compatibility (except in the newly introduced keycloak_uma client https://github.com/marcospereirampj/python-keycloak/pull/403).
* It deprecates all connection related properties in the keycloak_admin class.
* This PR removes the token parameter from all keycloak_uma client method signatures. The token is no longer necessary as the token is auto-refreshed as required. While this is technically a breaking change, the client was introduced recently enough that it's unlikely to have a wide impact and would make it much more user-friendly.
* The auto_refresh_token property has been deprecated and performs a no-op. Instead, the token is always automatically refreshed when it hits 90% of its lifetime. This makes it easier for users, as it is no longer necessary to decide on a per method basis when to refresh.
* In future, it would probably be preferable to simply accept the connection object as a constructor parameter for all clients, further separating concerns. The keycloak_uma client has already been changed to support this, and the keycloak_admin client now accepts a connection parameter to support this, but does not remove any existing parameters for backward compatibility.
* The deprecation package has been introduced as a dependency for orderly management of deprecations.
* The freezegun package has been introduced as a dev-dependency for freezing time during testing of token expiration.


nuwang:
The test failures are unrelated, and occur even on the master branch against the latest keycloak image. Tests pass locally against image `quay.io/keycloak/keycloak@sha256:b8f2a453a17a244a829fdafdb08dd77f719d3622bc3987c76a81771c0913b882` but fail against the very latest image.

ryshoooo:
Hi @nuwang 

This is amazing, thank you so much for this big refactoring :) I pretty much agree 100% with your points and overall your changes look fantastic. This is definitely a step in the right direction for this package :)

The tests should be fixed on master, if you can just resolve the conflicts on the lockfile, we're good to go :)

nuwang:
@ryshoooo Thanks for reviewing and glad you liked the changes. I've made a few more changes (last 3 commits), namely, updating the docs, and renaming some variables. In particular, KeycloakOpenIDConnectionManager was renamed to KeycloakOpenIDConnection to reduce typing.

src/keycloak/keycloak_admin.py
```diff
@@ -214,146 +215,254 @@ def connection(self, value):
...
     def custom_headers(self, value):
-        self._custom_headers = value
+        self.connection.custom_headers = value
+
+    @property
+    @deprecation.deprecated(
+        deprecated_in="2.13.0",
+        removed_in="3.0.0",
+        current_version=__version__,
+        details="Use the connection.auto_refresh_token property instead",
```
ryshoooo: This is not correct, right? The `connection` does not have the `auto_refresh_token` attribute. Maybe a message like `"Auto-refresh will be implicitly set for all requests"` instead?

src/keycloak/keycloak_admin.py
```diff
@@ -214,146 +215,254 @@ def connection(self, value):
...
+        :rtype: list
+        """
+        return self._auto_refresh_token
 
     @auto_refresh_token.setter
+    @deprecation.deprecated(
+        deprecated_in="2.13.0",
+        removed_in="3.0.0",
+        current_version=__version__,
+        details="Use the connection.custom_headers property instead",
```

ryshoooo (commented):
Just a small update on the deprecation message :)

src/keycloak/keycloak_admin.py
nuwang: Thanks for catching that! Fixed.

Issue #416, opened 2023-03-03
get_group_by_path is broken with many groups


tobiasge:
We have a few thousand groups in our Keycloak server. When using "get_group_by_path" all groups are fetched from the server and filtered locally which leads to a timeout errror.

The Keycloak API provides a method to directly load a group by its path. I will create a PR to change get_group_by_path  to use this API endpoint.

PR #417, opened 2023-03-03
fix: get_group_by_path uses Keycloak API to load

tobiasge:
Keycloak provides an API endpoint to directly load a group by its path. This commit changes get_group_by_path to not filter the groups locally and instead uses the API endpoint.

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -1155,36 +1155,22 @@ class KeycloakAdmin:
 
         return self.__fetch_all(url, query)
 
-    def get_group_by_path(self, path, search_in_subgroups=False):
+    def get_group_by_path(self, path):
         """Get group id based on name or path.
 
-        A straight name or path match with a top-level group will return first.
-        Subgroups are traversed, the first to match path (or name with path) is returned.
+        Returns full group details for a group defined by path
 
         GroupRepresentation
         https://www.keycloak.org/docs-api/18.0/rest-api/#_grouprepresentation
 
         :param path: group path
         :type path: str
-        :param search_in_subgroups: True if want search in the subgroups
-        :type search_in_subgroups: bool
         :return: Keycloak server response (GroupRepresentation)
         :rtype: dict
         """
-        groups = self.get_groups()
-
-        # TODO: Review this code is necessary
-        for group in groups:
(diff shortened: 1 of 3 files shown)


tobiasge:
Fix for #416

tobiasge:
The failing tests are not related to the changes made in this PR.
I think the problem is Keycloak 21 for which the Docker image now runs on Java 17 where Nashorn (Javascript engine) has been removed.

ryshoooo:
Hi @tobiasge

Thanks a lot for your contribution! :) Overall LGTM and I'm really happy that we're getting rid of this logic for subgroup searching and leaving it to the Keycloak Server to handle it. The tests should be good now with the latest changes in master, so I'll merge it right away :)

Galaxy102:
I'd just like to inform you that this PR changed the behaviour of the library call. Previously, when called with `path="/"`, it would return None. Now it raises a `KeycloakGetError` with status code 404. The previous behaviour can be restored by using the following snippet:
```
kc_group: dict | None
try:
    kc_group = self._keycloak.get_group_by_path(path="/" + group_name)
except KeycloakGetError as e:
    if e.response_code != 404:
        raise
    kc_group = None
```

In my case `group_name` could've been `""`, indicating "No group".

ryshoooo:
> I'd just like to inform you that this PR changed the behaviour of the library call. Previously, when called with `path="/"`, it would return None. Now it raises a `KeycloakGetError` with status code 404. The previous behaviour can be restored by using the following snippet:
> 
> ```
> kc_group: dict | None
> try:
>     kc_group = self._keycloak.get_group_by_path(path="/" + group_name)
> except KeycloakGetError as e:
>     if e.response_code != 404:
>         raise
>     kc_group = None
> ```
> 
> In my case `group_name` could've been `""`, indicating "No group".

Hi @Galaxy102 

Thanks, I've noticed it myself when a part of the application started to fail, due to this exactly :) So we've accidentally included a breaking change without stating so. Honestly, I think the behavior that is now out (i.e. raising KeycloakGetError) is the right way to go about it, as it is more consistent with the rest of the methods. If there are more and more issues reported about this, I'll add a note to the documentation stating that it's an accidental breaking change and the users are therefore advised to use version < X.Y.Z or update their application code :)

PR #420, opened 2023-03-08
Relax the poetry versions of core dependencies.

ari-nz:
Hi,

I'm using this library in one of my apps and getting version conflicts between some package versions. 
The cause for me seems to be related to how Poetry uses the `^` character for package versions. 

```
beepboop = "^2.1.7"
# Equivalent to >=2.1.7, <3.0.0
```

The issue I hope to fix in this PR is the removal of those upper bounds in the core dependencies.

That upper bound causes clashes with other packages that have other transitive dependencies. (an example shown below, taken from https://hynek.me/articles/semver-will-not-save-you/)
image

Pros:
* Extends the use of this package into situations and use cases, and stops people forking and duplicating the package to work for their use cases

Cons:
* A major version change in a dependency _may_ cause an error down the line (however if you are mainly using core functions of these packages this is less likely)

Should a breaking change to a dependency be found in the future, then a `<` condition should be added in future

Related:
Discussing in the core poetry repo: https://github.com/python-poetry/poetry/issues/2731
Version Spec documentation: https://python-poetry.org/docs/dependency-specification/


ryshoooo:
Hi @ari-nz 

This change looks reasonable and good :) Just make sure you follow the conventional commits guidelines for commit messages, then we should be good to go.

ryshoooo:
These changes have been included in the 2.16.6 release :)

PR #424, opened 2023-03-15
Feature/initial access token

mklassen:
Added support for obtaining an initial access token using `KeycloakAdmin` and creating a client with the initial access token using `KeycloakOpenID`.


ryshoooo:
Hi @mklassen 

Thanks a lot for your contribution :) LGTM!

PR #425, opened 2023-03-21
fix: deprecation warnings in keycloak_admin.py

### LINKED ISSUE #422: Deprecation warnings with 2.13.2 in admin module
After updating to 2.13.2, I keep getting these warnings:
```
/keycloak/keycloak_admin.py:1445: DeprecatedWarning: realm_name is deprecated as of 2.13.0 and will be removed in 3.0.0. Use the connection.realm_name property instead
    params_path = {"realm-name": self.realm_name, "id": client_id}
```
```
/keycloak/keycloak_admin.py:152: DeprecatedWarning: auto_refresh_token is deprecated as of 2.13.0 and will be removed in 3.0.0. Auto-refresh will be implicitly set for all requests
    self.auto_refresh_token = auto_refresh_token
```

pehala:
Fixes #422

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -149,7 +149,8 @@ class KeycloakAdmin:
             custom_headers=custom_headers,
             timeout=timeout,
         )
-        self.auto_refresh_token = auto_refresh_token
+        if auto_refresh_token is not None:
+            self.auto_refresh_token = auto_refresh_token
 
     @property
     @deprecation.deprecated(
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -545,7 +545,7 @@ class KeycloakAdmin:
         :rtype: dict
         """
         params_path = {
-            "realm-name": self.realm_name,
+            "realm-name": self.connection.realm_name,
             "export-clients": export_clients,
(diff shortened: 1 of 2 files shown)


ryshoooo:
Hi @pehala 

LGTM, except for the removal of `auto_refresh_token` attribute. I understand that it's not "needed", however removing it in this PR would introduce a breaking change. So there are 2 options:

1. Keep `auto_refresh_token` attribute in the KeycloakAdmin class, i.e. revert the deletion
2. Update the commit message to include a breaking change note as conventional commits specify

I'm more prone towards the option 1. The `python-keycloak` version 3.0.0 should have all of these deprecated attributes completely removed, not just one.

pehala:
I can revert the deletion. However, that would not solve the false deprecation warnings when using KeycloakAdmin. Is there a way to keep the attribute and have no deprecation warnings?

ryshoooo:
> I can revert the deletion. However, that would not solve the false deprecation warnings when using KeycloakAdmin. Is there a way to keep the attribute and have no deprecation warnings?

A solution could be something like

```
if auto_refresh_token is not None:
    self.auto_refresh_token = auto_refresh_token
```

pehala:
@ryshoooo Thanks, fixed

ryshoooo (approved):

PR #427, opened 2023-03-24
fix: bad typing

kunlabs:
fix a bad typing which make VSCode complaining.

`connection: KeycloakOpenIDConnection | None = None` is the correct typing also, but only works with python 3.10+ 
so I did use the `Optional`

スクリーンショット 2023-03-24 22 12 05

This library is bad typed in general, may better if we can to add some type check to improve it.


ryshoooo (approved):
Hi @ykun91 

Thanks a lot! :) LGTM

PR #429, opened 2023-04-06
fix: Check if _s exists in ConnectionManager before deleting it

### LINKED ISSUE #423: `KeycloakOpenIDConnection` object has no attribute '_s'
After updating to 2.13.2, I keep encountering this error. I do not call it directly, nor I am aware of me using it within any context manager
```
  Traceback (most recent call last):
    File "keycloak/connection.py", line 89, in __del__
      self._s.close()
  AttributeError: 'KeycloakOpenIDConnection' object has no attribute '_s'
```

pehala:
Simple but ugly fix. A better option would be to move the `__init__` call in `KeycloakOpenIDConnection`, which is impossible due to custom_headers.

Fixes #423

### DIFF
--- a/src/keycloak/connection.py
+++ b/src/keycloak/connection.py
@@ -86,7 +86,8 @@ class ConnectionManager(object):
 
     def __del__(self):
         """Del method."""
-        self._s.close()
+        if hasattr(self, "_s"):
+            self._s.close()
 
     @property
     def base_url(self):


ryshoooo:
LGTM. However, to be frank, I'm not sure why the `__del__` is overridden in the first place. Also based on mcoding's analysis (https://www.youtube.com/watch?v=IFjuQmlwXgU), `__del__`'s behavior is just really confusing and leads to unnecessary errors. I'd be fine if the entire `def __del__...` is removed :)

pehala:
I think it is because of session management. I believe not closing them could potentially leave the socket open on OS, until termination. But yeah `__del__` is a TERRIBLE way how to manage it

ryshoooo:
> I think it is because of session management. I believe not closing them could potentially leave the socket open on OS, until termination.

Right, I see. The request's Session does not close it upon destruction unless it's used as a context manager :) alright then

PR #444, opened 2023-05-12
Add query to get users group method and permit pagination

### LINKED ISSUE #443: The get_user_groups method of keycloak_admin doesn't allow paging, however in the Keycloak API it is available
Add a query dictionary so you can pass more optional parameters.

The method: https://github.com/marcospereirampj/python-keycloak/blob/e483bf4323c29e144448363bac23a157aa1d9328/src/keycloak/keycloak_admin.py#L870

Captura de Tela 2023-05-12 às 10 33 44

anapaulamendes:
Issue #443

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -867,24 +867,33 @@ class KeycloakAdmin:
         data_raw = self.connection.raw_get(urls_patterns.URL_ADMIN_USER.format(**params_path))
         return raise_error_from_response(data_raw, KeycloakGetError)
 
-    def get_user_groups(self, user_id, brief_representation=True):
+    def get_user_groups(self, user_id, query=None, brief_representation=True):
         """Get user groups.
 
         Returns a list of groups of which the user is a member
 
         :param user_id: User id
         :type user_id: str
+        :param query: Additional query options
+        :type query: dict
         :param brief_representation: whether to omit attributes in the response
         :type brief_representation: bool
         :return: user groups list
         :rtype: list
         """
+        query = query or {}
+
         params = {"briefRepresentation": brief_representation}
+        query.update(params)
+
         params_path = {"realm-name": self.connection.realm_name, "id": user_id}
-        data_raw = self.connection.raw_get(
-            urls_patterns.URL_ADMIN_USER_GROUPS.format(**params_path), **params
-        )
-        return raise_error_from_response(data_raw, KeycloakGetError)
+
+        url = urls_patterns.URL_ADMIN_USER_GROUPS.format(**params_path)
+
+        if "first" in query or "max" in query:
+            return self.__fetch_paginated(url, query)
+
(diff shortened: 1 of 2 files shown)


src/keycloak/keycloak_admin.py
```diff
@@ -867,24 +867,33 @@ def get_user(self, user_id):
...
         params_path = {"realm-name": self.connection.realm_name, "id": user_id}
-        data_raw = self.connection.raw_get(
-            urls_patterns.URL_ADMIN_USER_GROUPS.format(**params_path), **params
-        )
-        return raise_error_from_response(data_raw, KeycloakGetError)
+
+        url = urls_patterns.URL_ADMIN_USER_GROUPS.format(**params_path)
+
+        if "first" in query or "max" in query:
+            return self.__fetch_paginated(url, query)
```
ryshoooo, line 895: Could you add a unit test that makes sure the pagination works as expected? :)

anapaulamendes: Of course!

anapaulamendes: Done!

anapaulamendes: There is a pagination problem with "first" parameter only in keycloak version 20.0
I debugged the code to see where the error could be, but calling directly to keycloak 20.0 got the same problem. Can you help me @ryshoooo ?

anapaulamendes: @ryshoooo  it seems to me that the problem is really in version 20.0. I found this issue listed in the Keycloak 21.0.0 released:

Release notes: https://www.keycloak.org/2023/02/keycloak-2100-released.html
Issue [#16899](https://github.com/keycloak/keycloak/issues/16899)

ryshoooo: Right well, in some way, it's nice to know that our test framework catches Keycloak issues well :)

The assurance/claim we're giving to our users is that this package is meant to be working with the latest Keycloak version, so we don't necessarily have to be worried about older Keycloak versions for a new package release. The tests for previous Keycloak versions were also introduced in order to show that if the tests fail for the latest Keycloak but not for the older versions, the changes made in the PR are "OK", the tests are just not up-to-date with the latest Keycloak.

The correct way to resolve this, if we have tests depending on the Keycloak version to be >= X, is to have a decorator like

```python
@version(">=21")
def test_that_only_works_for_keycloak_21_or_above(...):
    ...
```

The decorator would check the Keycloak version and if the version does not satisfy the constraint, the test would be skipped.

It's not difficult to code such a decorator, but I'm also fine merging this with the tests failing and creating the decorator later :) @anapaulamendes Let me know whether you would like this to be merged or you want to have a stab at the decorator :)

anapaulamendes: You can merge. Thanks @ryshoooo!

ryshoooo (approved):

PR #446, opened 2023-05-14
Fixes `Authorization.load_config`  breaking when a scope based permission is linked with anything other than a role based policy

### LINKED ISSUE #445: `Authorization.load_config` breaks when authorization config contains scope based permission linked with anything other than a role based policy
Its pretty clear that the library only supports loading role based policies for now. When creating a scope based permission, we should first check if the policy linked to that permission has been loaded by us and if not, we should skip adding that policy to the `policies` attribute of the `Authorization` class. Since we're pretty much doing the same thing with resource based permissions, I'm not sure why we haven't done the same with scope based permissions.

sras1599:
Fixes #445


ryshoooo (approved):
Hi @sras1599 

Thanks a lot for the fix! :) LGTM

Issue #428, opened 2023-03-30
KeycloakAdmin realm_name gets overwritten with user_realm_name if using a client secret to log in


gboor:
When I initialize a `KeycloakAdmin` object with both a `realm_name` and a `user_realm_name`, using a client secret (service account) to log in, the returned instance has both `realm_name` and `user_realm_name` set to the original value of `user_realm_name`.

Example;

```
adm = KeycloakAdmin(server_url="https://blah.blah", client_id="admin-cli", client_secret_key="xxxyyy", user_realm_name="master", realm_name="test");

print(adm.connection.realm_name);  # outputs "master", not "test"
```

A workaround is manually setting `adm.connection.realm_name` to the correct realm name after the object is created, but that seems a bit double.

The reason for this problem is around line 300 in `openid_connection.py` where the following snippet lives;

```
        if self.client_secret_key:
            if self.user_realm_name:
                self.realm_name = self.user_realm_name
            grant_type.append("client_credentials")
```

Now, I understand the `grant_type` being set, but I really do not understand why the `realm_name` gets set to the `user_realm_name` when logging in with a service account...

Note that this also applies when using a confidential service to log in, which also requires the `client_secret_key` to be set. So this is not JUST limited to service accounts as far as I can see.

I can issue an MR to remove this, but I would like to understand why this is here in the first place to begin with...

ryshoooo:
Hi @gboor 

I agree with your analysis :) I don't see any point in overriding the `realm_name` with `user_realm_name`. The `user_realm_name` is used only for fetching the admin user token, it shouldn't just override the KeycloakAdmin's set realm name.

I also might not be seeing the use case behind this decision right now, but logic and consistency point to these 2 lines being deleted :) Feel free to issue the PR

PR #447, opened 2023-05-16
fix: do not swap realm for user_realm when logging in with a client service account

gboor:
Solves #428 by not overwriting the realm when logging in with a client_id and client_secret to a different user_realm.


ryshoooo (approved):
Thanks for the fix! :) LGTM

PR #449, opened 2023-05-26
enh: improve performance of get_user_id

mjeanroy:
This PR improves the performance of the `get_user_id` function by changing how to retrieve a user from a username:

- The previous implementation used the `users` endpoint by using the search query, which starts to be slow when you have thousands of users.
- The new implementation still uses the `users` endpoint, but it queries the endpoint by specifying the username parameter, a max of 1, and an exact, which is much faster.


ryshoooo (approved):
Hi @mjeanroy 

Thanks a lot for this change! :) Happy to see this coming in! LGTM

Issue #458, opened 2023-06-22
Double slash in url for IdP


jasminapegan:
I think this double slash in URL_ADMIN_IDP is a mistake, the url should read "admin/realms/{realm-name}/identity-provider/instances/{alias}" not "admin/realms//{realm-name}/identity-provider/instances/{alias}".
https://github.com/marcospereirampj/python-keycloak/blob/257de54a711350eefbe84765ee382a4ff91e9a27/src/keycloak/urls_patterns.py#L148

PR #459, opened 2023-06-22
fix: remove duplicate slash in URL_ADMIN_IDP

jasminapegan:
resolves issue
https://github.com/marcospereirampj/python-keycloak/issues/458


ryshoooo (approved):
Hi @jasminapegan 

Thanks a lot for the PR! This is indeed a mistake :)

PR #383, opened 2022-10-29
Implement missing admin method create_client_authz_scope_based_permission()

fschrempf:
This provides a solution for https://github.com/marcospereirampj/python-keycloak/issues/370. I plan to look into how to write a testcase for this, so I will leave this as a draft for now.

Issue #370, opened 2022-08-18
Add possibility to create scope-based permission


ouattararomuald:
**Request Type:**

- **Feature**

**Explanations:**

In `KeycloakAdmin` we have a method `create_client_authz_resource_based_permission` that allows to create resource-based permission of client.

Right now I'm developing an app just noticed, unless I missed it, that there's no method to create scope-based permission for a client.

**Request:**

Add a method `KeycloakAdmin.create_client_authz_scope_based_permission`.

PR #460, opened 2023-06-22
Implement missing admin method create_client_authz_scope_based_permission() and create_client_authz_policy()

cainotis:
This finishes #383 providing a solution to #370 and allows to create any type of policy


cainotis:
I believe that something went wrong with the test for Python 3.10 20.0, it didn't pass in the test_token() because the expire time differed by 1. I can't re-run the test without making another commit so I will wait for the reviewers before doing anything unnecessary.
Here is the difference appointed by the test:
  {'refresh_expires_in': 1799} != {'refresh_expires_in': 1800}
  {'expires_in': 299} != {'expires_in': 300}

src/keycloak/urls_patterns.py
```diff
@@ -121,6 +121,9 @@
 URL_ADMIN_CLIENT_AUTHZ_RESOURCE_BASED_PERMISSION = (
     URL_ADMIN_CLIENT_AUTHZ + "/permission/resource?max=-1"
 )
+URL_ADMIN_CLIENT_AUTHZ_SCOPE_BASED_PERMISSION = (
+    URL_ADMIN_CLIENT + "/authz/resource-server/permission/scope?max=-1"
```
ryshoooo: Can we utilize `URL_ADMIN_CLIENT_AUTHZ` instead of `URL_ADMIN_CLIENT`?

cainotis: Yes, I updated the file to use it

ryshoooo (approved):
Thanks a lot for your contribution :) LGTM!

PR #469, opened 2023-07-20
Remove removed function attribute from README.md

vitorespindola:
Remove option search_in_subgroups from get_group_by_path usage example on README. This option was removed on b3dd7d49046ff99d2af152d5a3bba1eb64eb92b7


ryshoooo (approved):
Indeed :) Thanks a lot @vitorespindola!

PR #474, opened 2023-08-03
Feat/support admin events

zimaldone:
Hi :), 
apologies for any mistake in the process.

A very small addition to support also Admin events and not only User events.
I also fixed a couple of tests based on the latest default settings of Keycloak 22.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,18 @@ keycloak_admin.realm_name = "demo" # Change realm to 'demo'
 keycloak_admin.get_users() # Get users in realm 'demo'
 keycloak_admin.create_user(...) # Creates a new user in 'demo'
 
+# Get User events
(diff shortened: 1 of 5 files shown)


zimaldone:
Hi @marcospereirampj & @ryshoooo ,
It's definitely not my intention to bother you with PR requests but I'd only like to understand if there's any way to get a feedback and possibly having the code merged.
I'm happily using the library and I'd not like to switch to a crafted/forked version I maintain, but I'd prefer to rely on the well developed code you have here.
Let me know if I can help for any clarification... I think the change I submitted is pretty simple and small.

Cheers and thanks in advance.

ryshoooo (approved):
Thanks a lot @zimaldone!

LGTM! :)

PR #478, opened 2023-08-18
feat: add KeycloakAdmin.get_idp()

derlin:
Add the missing `get_idp(alias)` method. Note that the failing tests are the one from master (related to auth), the idp ones finish with success.


derlin:
@ryshoooo, @marcospereirampj  any thoughts?

ryshoooo (approved):
Thanks a lot @derlin!

Very useful endpoint, I usually just went through the list given by `get_idps`, so this is a great improvement :)

Issue #479, opened 2023-08-24
Allow the verify attribute to be bool or string


waza-ari:
Both the `KeycloakOpenID` class as well as the `KeycloakAdmin` class (and subsequently also the `ConnectionManager ` and `KeycloakOpenIDConnection` support a `verify` parameter which defines whether the SSL certificate should be checked. Currently, this is specified as being a boolean value. 

That parameter is then subsequently passed to `requests` `Session` methods like `get` and `put`. According to the `requests` [API description,](https://requests.readthedocs.io/en/latest/api/) this verify parameter actually accepts a boolean value or a string, in which case the value must be a path to a custom CA file used for certificate validation.

Passing a string works already today as the parameter is not validated and passed straight to the `requests` methods. I suggest updating the documentation and make clear that both a boolean value or a string can be passed.

This discussion came up during a MR on my own library making use of yours. See the [following discussion](https://github.com/waza-ari/fastapi-keycloak-middleware/pull/12) for more details.

PR #480, opened 2023-08-24
Allow the verify attribute to be bool or string #479

waza-ari:
This PR only updates docs as requested in #479


ryshoooo (approved):
Thanks a lot @waza-ari!

LGTM :)

PR #483, opened 2023-08-31
_version.py no longer uses pkg_resources to retrieve the version

pojzn:
Switched from pkg_resources to importlib due to deprecation warnings.

### DIFF
--- a/src/keycloak/_version.py
+++ b/src/keycloak/_version.py
@@ -21,6 +21,6 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-import pkg_resources
+from importlib import metadata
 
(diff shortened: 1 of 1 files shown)


src/keycloak/_version.py
```diff
@@ -21,6 +21,6 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-import pkg_resources
+from importlib import metadata
```
fangchenli, line 24: This is only available for Python 3.8+. So you need to add a compat shim for Python 3.7.

fangchenli (changes requested):

src/keycloak/_version.py
ryshoooo: We're dropping support for Python 3.7, as Python 3.7 itself is not supported by Python https://devguide.python.org/versions/, so this is fine :)

ryshoooo:

PR #491, opened 2023-10-19
Update dynamic client using registration access token

### LINKED ISSUE #431: keycloak_openid attribute not set in KeycloakOpenIDConnection
If the `token` argument is passed during the initialization of a `KeycloakOpenIDConnection` instance, then the `get_token()` method is not called. However when the `refresh_token()` is subsequently called it attempts to access the `self.keycloak_openid` attribute which is set in `get_token()`. This results in the following error:

```
AttributeError: 'KeycloakOpenIDConnection' object has no attribute 'keycloak_openid'
```

The `keycloak_openid` should be an `@property` that creates the required `KeycloakOpenID` as required by either the `get_token()` or `refresh_token()` methods. It could also be an `@cached_property` to automatically handle the caching.

mklassen:
- Added support for dynamic client update using registration access token.
- Fix #431 error for missing keycloak_openid
- Remove internal use of some deprecated methods
- Modified tests to accommodate changes to authenticator providers and flows in keycloak 22.0

### DIFF
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -41,6 +41,7 @@ from .exceptions import (
     KeycloakGetError,
     KeycloakInvalidTokenError,
     KeycloakPostError,
+    KeycloakPutError,
     KeycloakRPTNotFound,
     raise_error_from_response,
 )
@@ -49,6 +50,7 @@ from .urls_patterns import (
     URL_AUTH,
     URL_CERTS,
     URL_CLIENT_REGISTRATION,
+    URL_CLIENT_UPDATE,
     URL_ENTITLEMENT,
     URL_INTROSPECT,
     URL_LOGOUT,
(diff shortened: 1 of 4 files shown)


src/keycloak/keycloak_openid.py
```diff
@@ -711,3 +713,31 @@ def register_client(self, token: str, payload: dict):
             URL_CLIENT_REGISTRATION.format(**params_path), data=json.dumps(payload)
         )
         return raise_error_from_response(data_raw, KeycloakPostError)
+
+    def update_client(self, token: str, client_id: str, payload: dict):
```
ryshoooo, line 717: Shouldn't this be in the KeycloakAdmin class instead?

mklassen: This is to update dynamically created clients using a registration access token. Typically for a client previously dynamically created with an initial access token. This allows OpenID client applications to create and update their own keycloak client without having admin access to keycloak. Since dynamic client registration is part of the OpenID specification (https://openid.net/specs/openid-connect-registration-1_0.html), I think this belongs in the OpenID client.

ryshoooo (approved):
Thanks a lot @mklassen!

LGTM :)

PR #493, opened 2023-10-24
fix: remove python3.7 from CI (because it's unsupported since june 2023)

Kevin-Bannier:
Hello,

Since june 2023, Python 3.7 has been declared unsupported (https://devguide.python.org/versions/#unsupported-versions)

This PR remove it from CI and add the version Python 3.12 instead


ryshoooo:

PR #496, opened 2023-10-26
fix: Add typing to Keycloak admin connection

jackylamhk:


src/keycloak/keycloak_admin.py
```diff
@@ -84,7 +84,7 @@ class KeycloakAdmin:
     PAGE_SIZE = 100
 
     _auto_refresh_token = None
-    _connection = None
+    _connection: Union[KeycloakOpenIDConnection, None] = None
```
krisfremen: Should this align with [Optional[KeycloakOpenIDConnection]](https://github.com/marcospereirampj/python-keycloak/pull/496/files#diff-b4a20b26ca73014f6817549eb060b7eda594e199a8c7a700f338bff93ba1e6abR104) or vice versa?

jackylamhk: You're right - let me change that.

ryshoooo (approved):
Thanks a lot for your contribution @jackylamhk!

LGTM :)

PR #499, opened 2023-10-31
make audience in token exchange optional

### LINKED ISSUE #498: Audience should be optional in Token Exchange
The `exchange_token` method expects an `audience` to be passed. According to the [referenced documentation](https://access.redhat.com/documentation/en-us/red_hat_single_sign-on/7.2/html/securing_applications_and_services_guide/token-exchange#doc-wrapper) it is optional. For example, it is not needed for an impersonation.

pitwegner:
refs #498


ryshoooo (approved):
LGTM! Thanks a lot for your contribution @pitwegner!

PR #500, opened 2023-11-02
chore: test against 3.12 python

krisfremen:
Adds Python 3.12 to the testing matrix.


ryshoooo:

PR #505, opened 2023-11-13
fix: Add direct "setuptools" dependency

pehala:
With Python 3.12 the `setuptools` is no longer enabled by default, which makes python-keycloak fail because of `pkg_resources`.
Similarly with Poetry 1.7.0, the `poetry install --sync` also removes the `setuptools`, which results in the same error

I am not sure which version is reasonable to depend on, so I added any for now.


ryshoooo (approved):
Thanks a lot @pehala!

I'll probably set a lower bound for the version a bit later, but this is good enough for now :)

PR #506, opened 2023-11-13
Ci/fix tests

ryshoooo:
Fixed tests, dropped Python 3.7, updated dependencies, sanitized a bit the latest PR contributions.


pyproject.toml
```diff
@@ -29,20 +29,20 @@ Documentation = "https://python-keycloak.readthedocs.io/en/latest/"
 "Issue tracker" = "https://github.com/marcospereirampj/python-keycloak/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.7,<4.0"
-setuptools = "*"
+python = ">=3.8,<4.0"
+setuptools = ">=50.0.0"
```
pehala: Is it still needed if you switched to `importlib`, not sure if it is used somewhere else too?

ryshoooo: True, I don't think it is, at least nothing was failing when it was removed during my testing. Thanks, will remove :)

ryshoooo: Looks like `babel` has `setuptools` included anyway. But yeah, we don't explicitly need it. Removed

PR #512, opened 2024-01-04
feat: allows retrieval of realm and client level roles for a user

comemaryvallee:
Adding method **get_all_roles_of_user** to KeycloakAdmin.
Adding **URL_ADMIN_USER_ALL_ROLES** to url_patterns

This PR adds the functionality of retrieving roles from realm and client levels for a user.
Within my team, we need to retrieve all of a user's clients roles.
Currently, the only way to do this is by looping through all clients in a realm and using the KeycloakAdmin.get_client_roles_of_user function. This method requires a lot of calls and is very time consuming in the case of a large number of clients.
Using the [GET /admin/realms/{realm}/users/{id}/role-mappings](https://www.keycloak.org/docs-api/22.0.1/rest-api/index.html#_get_adminrealmsrealmusersidrole_mappings) endpoint of the Keycloak REST API would meet our need with only one request.

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -3000,6 +3000,20 @@ class KeycloakAdmin:
         )
         return raise_error_from_response(data_raw, KeycloakDeleteError, expected_codes=[204])
 
+    def get_all_roles_of_user(self, user_id):
+        """Get all level roles for a user.
+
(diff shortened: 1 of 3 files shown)


ryshoooo (approved):
Thanks a lot @comemaryvallee LGTM :)

PR #514, opened 2024-01-11
fix: use grant type password with client secret

mgm-jenstrinh:
When both user credentials and client credentials are supplied, this library used the grant_type=client_credentials. This fix will instead retrieve the token with grant_type=password, similar to the Keycloak admin CLI kcadm.sh.

### DIFF
--- a/src/keycloak/openid_connection.py
+++ b/src/keycloak/openid_connection.py
@@ -311,10 +311,10 @@ class KeycloakOpenIDConnection(ConnectionManager):
         The admin token is then set in the `token` attribute.
         """
         grant_type = []
-        if self.client_secret_key:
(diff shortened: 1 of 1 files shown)


mgm-jenstrinh:
The access token retrieved via grant_type=client_credentials has fewer privileges than the token retrieved with grant_type=password. This is important if I have an admin user and the admin-cli client is using client authentication.

For confidential clients, the resource owner password grant allows the additional specification of a client secret https://datatracker.ietf.org/doc/html/rfc6749#section-4.3. In contrast, the client credentials grant does not process any username and password parameter. Therefore it makes sense that grant_type=password is used when all three parameters (secret, username, password) are specified.

PR #517, opened 2024-01-22
fix: name of client_id parameter

tobiasge:
The name `client_role_id` for this parameter was confusing because it expected the client ID.


tobiasge:
Failing tests are caused by the changes to the group endpoints in Keycloak 23.

marcospereirampj (approved):

PR #534, opened 2024-03-15
feat: Allow query parameters for group children

tobiasge:
Allow query parameters on `get_group_children` for pagination and briefRepresentation.

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -1349,21 +1349,24 @@ class KeycloakAdmin:
         # went through the tree without hits
         return None
 
-    def get_group_children(self, group_id):
-        """Get group children by id.
+    def get_group_children(self, group_id, query=None):
+        """Get group children by parent id.
 
         Returns full group children details
 
-        :param group_id: The group id
+        :param group_id: The parent group id
         :type group_id: str
+        :param query: Additional query options
+        :type query: dict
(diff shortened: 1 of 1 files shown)


src/keycloak/keycloak_admin.py
```diff
@@ -1349,21 +1349,24 @@ def get_subgroups(self, group, path):
         # went through the tree without hits
         return None
 
-    def get_group_children(self, group_id):
-        """Get group children by id.
+    def get_group_children(self, parent_group_id, query=None):
```
ryshoooo: I agree with the naming, unfortunately, this change would make this a breaking change for users relying on the argument being named `group_id`, thus making an upgrade for them to break the working functionality.

Python-keycloak version 4.X will come with a couple more breaking changes than just this one. I'd suggest either keeping the argument name the same (group_id) or creating a new method and adding a deprecated warning to this one.

I'm also open to hearing other suggestions of how to resolve this (perhaps merging this into release/4-like branch, etc.)

ryshoooo (changes requested):

src/keycloak/keycloak_admin.py
tobiasge: I will change the parameter name back to group_id.

ryshoooo (approved):
Thanks a lot @tobiasge :) LGTM

PR #535, opened 2024-03-15
fix: Tests with Keycloak 24

tobiasge:
Updates the error regex expressions for Keycloak 24.


ryshoooo (approved):
Thanks a lot @tobiasge!

Issue #537, opened 2024-03-20
 Issue with get_groups method pagination not working at all 


kernelshard:
Hi there,

I'm experiencing an issue while using  the pagination of results when using the get_groups method to retrieve groups from a Keycloak realm.

**Issue Description:**
When calling the get_groups method with parameters for "first" and "max" to limit the number of results, it seems that these parameters are not being respected. Instead, the method returns all the groups for the realm, regardless of the specified pagination values.

Steps to Reproduce:

Instantiate a KeycloakAdmin object with the appropriate credentials.
Call the get_groups method with parameters for "first" and "max".
Observe that all groups for the realm are returned, instead of the expected subset based on pagination parameters.
Expected Behavior:
The get_groups method should honour the "first" and "max" parameters, returning only the specified number of groups from the realm.

Actual Behavior:
The method returns all groups for the realm, regardless of the pagination parameters provided.

Code Snippet:

```
from keycloak import KeycloakAdmin

keycloak_admin = KeycloakAdmin(
    server_url="http://localhost:8080/auth",
    realm_name="master",
    username="admin",
    password="admin",
    client_id="id_of_the_client",
    client_secret_key="secret"
)

groups = keycloak_admin.get_groups({"first": 1, "max": 2})

```

Additional Information:
python-keycloak library version: 3.9.1
Python version: [3.11.7]
Docker Image: [quay.io/keycloak/keycloak:latest]
I would appreciate any assistance or guidance on resolving this issue. Thank you for your attention.

PR #541, opened 2024-03-21
fix: get_groups pagination not working #537

kernelshard:
Issue Link: https://github.com/marcospereirampj/python-keycloak/issues/537
Why: 
When calling the get_groups method with parameters for "first" and "max" to limit the number of results, it seems that these parameters are not being respected. Instead, the method returns all the groups for the realm, regardless of the specified pagination values.

What:
normal call was modifying pagination call everytime.


ryshoooo (approved):
Thanks a lot @samims :) LGTM

Issue #544, opened 2024-04-08
[Keycloak Admin.create_user() ] Return Type Hint is incorrect


fabinca:
![Screenshot from 2024-04-08 15-07-32](https://github.com/marcospereirampj/python-keycloak/assets/73344238/b40a7cde-89fd-4ac2-b842-d6bd5db1662a)
![Screenshot from 2024-04-08 15-08-07](https://github.com/marcospereirampj/python-keycloak/assets/73344238/c3f92732-f944-459c-a793-5ba8ee16e2f2)

```
user_id = keycloak_admin.create_user(
        {
            "username": username,
            "enabled": True,
            "emailVerified": True,
            "credentials": [{"value": password, "type": "password"}],
        },
        exist_ok=True,
    )
```

Above function returns the user_id, however type hints suggest it returns a dict, documentation suggests it returns a user representation.

ryshoooo:
Thanks for the report, I believe this is a duplicate of https://github.com/marcospereirampj/python-keycloak/issues/421

fabinca:
true I'm working on the pr

Issue #421, opened 2023-03-09
`KeycloakAdmin.create_user` provide wrong return type hint


icycandle:
https://github.com/marcospereirampj/python-keycloak/blob/9af82647ee9d940d3951de241f36ecdbbbfda93a/src/keycloak/keycloak_admin.py#L786

inline document says it return UserRepresentation dict, but the code infact return user_id str.

PR #545, opened 2024-04-08
fix: keycloak_admin.create_user documentation/ typehint

fabinca:
the type hint / return value was incorrect, the method returns user_id, of type: str 
here is the fix for this.

refers to issue: 
https://github.com/marcospereirampj/python-keycloak/issues/421
https://github.com/marcospereirampj/python-keycloak/issues/544


ryshoooo (approved):
Thanks a lot @fabinca ❤️ LGTM

PR #546, opened 2024-04-08
docs: Update all Keycloak links

tobiasge:
Update all links to the Keycloak Admin REST API for the current version.

To get a list of all versions present in the doc-strings i used this:
```bash
grep -h "docs-api" src/keycloak/*py | sed 's/^[[:space:]]*//g' | cut -d '/' -f 5 | sort -u
```


ryshoooo (approved):
Thanks a lot @tobiasge! :) I'm assuming most of the links work in the latest version, so LGTM :)

PR #562, opened 2024-05-21
Updated KeycloakOpenID documentation about server_url

JosueFunez:
For KeyCloak 17+ the authentication url changed, is not necessary to add "/auth" at the end of the url because this will throw a 404 error.

### DIFF
--- a/docs/source/modules/openid_client.rst
+++ b/docs/source/modules/openid_client.rst
@@ -12,6 +12,7 @@ Configure client OpenID
     from keycloak import KeycloakOpenID
 
     # Configure client
+    # For KeyCloak 17+ the server url must be something like "http://localhost:8080" without "/auth"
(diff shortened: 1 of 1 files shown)


docs/source/modules/openid_client.rst
```diff
@@ -12,6 +12,7 @@ Configure client OpenID
     from keycloak import KeycloakOpenID
 
     # Configure client
+    # For KeyCloak 17+ the server url must be something like "http://localhost:8080" without "/auth"
```
ryshoooo, line 15: Perhaps we can switch it around and set the example call without `/auth` and add a comment here that for versions older than 18, `/auth` is needed. Mainly as we don't guarantee these old versions to work with this library anyway, so the examples should work at least for the latest 5 versions of Keycloak.

ryshoooo (commented):
Thanks a lot @JosueFunez 

I've added just a single comment :) otherwise LGTM

PR #563, opened 2024-05-22
docs: example update

JosueFunez:
Example updated because for KeyCloak 17+ the authentication url changed. The "/auth" at the end of the server_url is not necessary; if we add this it will generate a 404 error.


ryshoooo (approved):
thanks a lot @JosueFunez :)

Issue #337, opened 2022-06-07
async version


Stasyanz:
any info about async version of the package?

ryshoooo:
Hi @Stasyanz 

Currently, we support neither any async version nor async functionality. My guess is that it'd take quite a significant amount of work to have this feature. And for that, I'd like to see and hear how many more are interested in this.

You can try patching `requests` with `requests-async` upon import, but I wouldn't really raise my hopes up too much. There are many places where we expect a response instead of a future response, but it could work for some methods.

obi-081:
Hi !
I'm interested too, it would be a very cool feature.
You may have a look at httpx if you want to do that, I think it is the best option.
https://www.python-httpx.org/
This would allow our WebServers to continue to serve requests when waiting for KeyCloak Response.

alexisgaziello:
It looks like there is an unmaintained forked repo that attempts to add the async functionality:

https://github.com/jegork/python-keycloak

Jdsleppy:
I have a handwritten async Keycloak client that I'm considering bundling up into a library for reuse.  I would prefer to put that effort into adding async to this library if the maintainers are interested in reviewing, merging, and releasing that work.

I would end up duplicating the existing test suite functions to make async versions of each test that could run against the same Keycloak instance.  As other commenters have suggested, I would consider moving from `requests` to `httpx` which has a nearly perfectly `requests`-compatible sync API, so hopefully most of the changes to existing code would just be replacing `requests` with `httpx`.  If that ends up being too difficult, then the new code could just use `httpx` and leave the sync version using `requests`.

Please let me know if you would be interested in incorporating changes like this into the library if I forked and developed it.  @ryshoooo @marcospereirampj

ryshoooo:
Hi @Jdsleppy 

I'm definitely interested in having async functionality as a native part of the library. So definitely feel free to create a PR :) I promise to find the time for review and merge.

drikster80:
> I have a handwritten async Keycloak client that I'm considering bundling up into a library for reuse. I would prefer to put that effort into adding async to this library if the maintainers are interested in reviewing, merging, and releasing that work.
> 
[...]

PR #565, opened 2024-05-29
Async feature

derekpy3ai:
Add support for async APIs using httpx client without interfering with the original functionality.

Linked issue: https://github.com/marcospereirampj/python-keycloak/issues/337


pyproject.toml
ryshoooo, line 1: I believe we need `httpx` as a direct dependency under `[tool.poetry.dependencies]`

src/keycloak/openid_connection.py
```diff
@@ -418,3 +419,71 @@ def raw_delete(self, *args, **kwargs):
...
+        and try *get* once more.
+
+        :param args: Additional arguments
+        :type args: tuple
+        :param kwargs: Additional keyword arguments
+        :type kwargs: dict
+        :returns: Response
+        :rtype: Response
+        """
+        self._refresh_if_required()
```
ryshoooo, line 436: the `_refresh_if_required` potentially calls `refresh_token`, which is blocking. I believe these methods need to become async as well in order to not introduce potential synchronous thread blocks

ryshoooo (commented):
Hi @pyaiman 

Wow, what a PR :) Thanks a lot for doing the hard work. We're definitely breaking the DRY principle here, but that's likely warranted due to the syntax of async/await.

The only worry I have is future divergence, i.e. when new methods are created in the sync version and they are forgotten to be added for the async version. Could we perhaps enforce it by a test that verifies that each sync method in each client has its async counterpart present? And vice versa. Also, not just the methods, but their signatures as well.

PR #569, opened 2024-06-07
docs: remove dependency lists from install.rst

Yelinz:
might make sense to remove the list? otherwise it gets forgotten when changing dependencies


ryshoooo:
> might make sense to remove the list? otherwise it gets forgotten when changing dependencies

Hi @Yelinz 

I'm completely fine with removing the list :)

Yelinz:
Removed, makes it easier to maintain :tada:

ryshoooo (approved):

PR #582, opened 2024-07-24
Retrieve the public key asynchronously in a_decode_token

kotnikd3:
Retrieve the public key asynchronously in a_decode_token()


ryshoooo (approved):
thanks a lot @kotnikd3 :)

PR #583, opened 2024-07-24
feat: allow the use of client certificates in all connections

nehuengonzalez:
Passing through cert attribute to requests and/or httpx.AsyncHTTPTransport in order to handle requests with client certificates.


nehuengonzalez:
I will check all changes and re-do the pull-request. Sorry for the inconvenience.

PR #584, opened 2024-07-29
feat: allow the use of client certificates in all requests

nehuengonzalez:
Passing through cert attribute to requests and/or httpx.AsyncHTTPTransport in order to handle requests with client certificates.

### DIFF
--- a/src/keycloak/connection.py
+++ b/src/keycloak/connection.py
@@ -49,9 +49,13 @@ class ConnectionManager(object):
     :type verify: Union[bool,str]
     :param proxies: The proxies servers requests is sent by.
     :type proxies: dict
+    :param cert: An SSL certificate used by the requested host to authenticate the client.
+        Either a path to an SSL certificate file, or two-tuple of
+        (certificate file, key file).
+    :type cert: Union[str,Tuple[str,str]]
     """
 
-    def __init__(self, base_url, headers={}, timeout=60, verify=True, proxies=None):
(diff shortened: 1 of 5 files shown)


src/keycloak/connection.py
```diff
@@ -65,11 +69,16 @@ def __init__(self, base_url, headers={}, timeout=60, verify=True, proxies=None):
...
+            Either a path to an SSL certificate file, or two-tuple of
+            (certificate file, key file).
+        :type cert: Union[str,Tuple[str,str]]
         """
         self.base_url = base_url
         self.headers = headers
         self.timeout = timeout
         self.verify = verify
+        self.cert = cert
         self._s = requests.Session()
```
ryshoooo, line 82: Doesn't `requests.Session` contain a field for `cert` already? Then you don't have to pass it in all the GET, POST, PUT, etc. methods

nehuengonzalez: Yes, absolutely. But the same happens with the verify attribute. Since the verify attribute was being passed through to each request I decided to do the same thing just to to be consistent. Nevertheless, I agree that it would be better to use Session. 

Should I modify the code and re-submit ?

Thank you for checking this!

ryshoooo: ah that's fair :) nah then let's keep it consistent, this is fine :)

ryshoooo:
Thanks a lot @nehuengonzalez LGTM :)

nehuengonzalez:
Thank you!

PR #585, opened 2024-07-31
5729 python keycloak

Khanindoc:
[5729-python-keyloak] remove old github action workflows and update patch version of pyproject.toml


ryshoooo:
Hi @Khanindoc 

What is the purpose of this PR? As it stands right now, I can't approve it nor merge it as it removes non-async HTTP client and all the pipelines. So I'm a bit wondering as to what is the aim here?

PR #592, opened 2024-09-13
feat: add matchingUri support for listing resources with wildcards

### LINKED ISSUE #587: Add matchingURI option for  UMAs resource_set_list_ids
Keycloak UMA Protection API provides the query option [matchingUri](https://github.com/keycloak/keycloak/blob/1f35877a107cf0855f0be4514b323f8b9fd0e867/services/src/main/java/org/keycloak/authorization/admin/ResourceSetService.java#L401) to query available resources in the form of e.g.

`
http://${host}:${port}/realms/${realm_name}/authz/protection/resource_set?uri=/items/1&matchingURI=true
` 
I.e. if clients Resource URIs in Admin Console are defined as "/items*", with this option set to "true" we can query with url="/items/1" and still get the resource name.

python-keycloak [resource_set_list_ids](https://github.com/marcospereirampj/python-keycloak/blob/76d2340db5a9baba8298439714fd0a98ca75ad08/src/keycloak/keycloak_uma.py#L205) doesn't support this option, so please add it.

Tarick:
Fixes #587 
Adds matchingUri boolean option for resource_set_list_id and a_resource_set_list_id to query for resources with URIs with wildcards.

### DIFF
--- a/src/keycloak/keycloak_uma.py
+++ b/src/keycloak/keycloak_uma.py
@@ -210,6 +210,7 @@ class KeycloakUMA:
         owner: str = "",
         resource_type: str = "",
         scope: str = "",
+        matchingUri: bool = False,
         first: int = 0,
         maximum: int = -1,
     ):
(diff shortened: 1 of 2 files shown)


ryshoooo (approved):
Thanks a lot @Tarick LGTM! :)

PR #596, opened 2024-09-26
feat: add ability to remove composite client roles

Syzygeezus:
Howdy!

This PR adds the ability to remove composite roles from client roles as per the below endpoint from the Keycloak Admin API docs:
`DELETE /admin/realms/{realm}/clients/{id}/roles/{role-name}/composites`

Thanks for the awesome package!


ryshoooo (approved):
Thanks a lot for the contribution @Syzygeezus! LGTM :)

PR #597, opened 2024-09-30
Set client_credentials as grant_type also when x509 certificate is given

alexrohozneanu:
When an x509 certificate is given to access the Keycloak admin interface we cannot get a new token. This PR fixes the issue

### DIFF
--- a/src/keycloak/openid_connection.py
+++ b/src/keycloak/openid_connection.py
@@ -43,6 +43,7 @@ class KeycloakOpenIDConnection(ConnectionManager):
     """
 
     _server_url = None
+    _grant_type = None
     _username = None
     _password = None
     _totp = None
@@ -59,6 +60,7 @@ class KeycloakOpenIDConnection(ConnectionManager):
     def __init__(
         self,
         server_url,
+        grant_type=None,
         username=None,
         password=None,
         token=None,
@@ -76,6 +78,8 @@ class KeycloakOpenIDConnection(ConnectionManager):
 
         :param server_url: Keycloak server url
         :type server_url: str
+        :param grant_type: grant type for authn
+        :type grant_type: str
         :param username: admin username
         :type username: str
         :param password: admin password
@@ -110,6 +114,7 @@ class KeycloakOpenIDConnection(ConnectionManager):
         self.token_lifetime_fraction = 0.9
         self.headers = {}
         self.server_url = server_url
+        self.grant_type = grant_type
         self.username = username
         self.password = password
         self.token = token
@@ -123,6 +128,12 @@ class KeycloakOpenIDConnection(ConnectionManager):
         self.custom_headers = custom_headers
         self.headers = {**self.headers, "Content-Type": "application/json"}
         self.cert = cert
+
+        if not self.grant_type:
+            if username and password:
+                self.grant_type = "password"
+            elif client_secret_key:
+                self.grant_type = "client_credentials"
 
         super().__init__(
             base_url=self.server_url,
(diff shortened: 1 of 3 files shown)


ryshoooo:
Hi @alexrohozneanu 

I'm failing to understand your PR. Why would we be setting the grant type to client_credentials if the connection to Keycloak is using private certificate? AFAIK, these are 2 orthogonal properties independent of each other, I can easily see the need for custom certificate without setting the client_credentials as the grant type, I just have my keycloak instance served with a custom CA.

Can you explain a bit more of what this PR is doing in case I'm misunderstanding it?

alexrohozneanu:
Hi @ryshoooo,

What I am talking about is mTLS.
I have created a client in Keycloak, and I've set the client authenticator as being x509 Certificate, thus the client_secret_key is not required.
When I try to issue a token using curl for example, the grant_type required is 'client_credentials' and I specify the client_id together with the certificate to be checked by Keycloak truststore.

```
curl -v -s --request POST 'https://localhost:8443/realms/master/protocol/openid-connect/token' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'client_id=some_client'   --data-urlencode 'grant_type=client_credentials' \
  --cert keycloak/1006.pem \
  --key keycloak/Keycloak.key \
--cacert keycloak/1001.pem
```

Regarding the library

I have something like this:

```python
self.__admin__ = LibKeycloakAdmin(
            server_url=kc_conf.get("server_url"),
            realm_name=kc_conf.get("realm_name"),
            client_id=kc_conf.get("client_id_admin"),
            cert={path_to_cert_and_key}
        )

self.__admin__.get_server_info()
```

image

The get_token() function ends up setting `self.token = None`, my get server info request failing

Let me know if it makes sense.

Thanks

ryshoooo:
Thanks for the details @alexrohozneanu :)

I think I understand now, you would like to use the client_credentials grant type with the X509 certificate authenticator, i.e. this flow https://www.keycloak.org/docs/latest/server_admin/#_x509?

In that case, I'd suggest creating a separate attribute on the KeycloakOpenID class instead of reusing `cert`, f.e. an optional `grant_type` attribute, which if specified, sets the grant_type in the `get_token` (and `a_get_token`) function to the given value. This is mainly because the use of `cert` is multifold, it can be used as in your case an authentication mechanism to the client, but it can be used just as a certificate for communicating with Keycloak over TLS. So it can't be used to auto-detect grant_type, because we just don't know what the purpose of the `cert` is for the user.

alexrohozneanu:
That's exactly what I am trying to do, @ryshoooo.

I can create an optional `grant_type` attribute, for sure.

However, I don't understand why the `grant_type` is created as a list in `get_token` (and `a_get_token`) function since we cannot give multiple grant_types when requesting a token because we will get a `duplicated_parameter` error. `{"error":"invalid_request","error_description":"duplicated parameter"}`

Can you give me more details in case I am missing something?

ryshoooo:
The grant_type as a list parameter with a single element has been here for a long time, before I got involved with this project :). It should not be list, it is just a mere "coincidence" that a list data form with single element gets converted into the same data structure as a key-value element for formurlencode application types. Feel free to make a `str` only :)

src/keycloak/openid_connection.py
```diff
@@ -110,6 +114,7 @@ def __init__(
         self.token_lifetime_fraction = 0.9
         self.headers = {}
         self.server_url = server_url
+        self.grant_type = grant_type
```
ryshoooo, line 117: Can we also add a getter and a setter for consistency?

ryshoooo (commented):
LGTM :) just one comment

ryshoooo (approved):

PR #599, opened 2024-10-04
feat: add scopes to device auth

marlenekoh:
Would like to add custom scopes when using keycloak's device auth code flow


ryshoooo (approved):
Thanks a lot @marlenekoh LGTM :)

PR #601, opened 2024-10-04
fix: changed keycloak admin unittests so they accomodate with new key…

alexrohozneanu:
…cloak release 26


ryshoooo (approved):
Thanks a lot @alexrohozneanu

PR #615, opened 2024-11-12
Make sure to not call sync IO functions inside async functions

Krismix1:
I found that several sync IO functions are used inside of `async` functions. This is not good, as sync IO may block the eventloop for undetermined amount of time.
I had to add a new `_a_token_info` that mimics `_token_info`; the only difference is that it calls the async alternatives inside.
I think those should be all the places where a sync IO was made.

I also found another discrepancy between `requests` and `httpx` - `httpx` does not handle `set`s for form-data requests.
Encoding in [requests](https://github.com/psf/requests/blob/23540c93cac97c763fe59e843a08fa2825aa80fd/src/requests/models.py#L119) vs [httpx](https://github.com/encode/httpx/blob/1805ee0d22f96e2b29f0500dd3076e10020004bf/httpx/_content.py#L141). httpx doesn't mention anything about support or lack of it for sets. Perhaps it could be reasonable for it to be fixed upstream.
Luckily some tests for `a_uma_permissions` caught this.

For now I fixed the problem inside the function itself.
The only other place I see where this could need a similar fix is `KeycloakOpenID.a_token`, which takes a user-supplied dict (the `extra`) and passes it as is to `httpx`.

### DIFF
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -892,6 +892,25 @@ class KeycloakOpenID:
         )
         return raise_error_from_response(data_raw, KeycloakPutError)
 
+    async def _a_token_info(self, token, method_token_info, **kwargs):
+        """Asynchronous getter for the token data.
+
+        :param token: Token
+        :type token: str
+        :param method_token_info: Token info method to use
+        :type method_token_info: str
+        :param kwargs: Additional keyword arguments passed to the decode_token method
+        :type kwargs: dict
+        :returns: Token info
+        :rtype: dict
+        """
+        if method_token_info == "introspect":
+            token_info = await self.a_introspect(token)
+        else:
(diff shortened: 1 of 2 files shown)


src/keycloak/keycloak_openid.py
```diff
@@ -689,7 +689,7 @@ def get_permissions(self, token, method_token_info="introspect", **kwargs):
         """
         if not self.authorization.policies:
             raise KeycloakAuthorizationConfigError(
-                "Keycloak settings not found. Load Authorization Keycloak settings ."
+                "Keycloak settings not found. Load Authorization Keycloak settings."
```
Krismix1, line 692: I hope this won't be a breaking change :D
Can revert it if you want to play it safe

ryshoooo (approved):
Thanks a lot @Krismix1 LGTM

PR #568, opened 2024-06-04
feat: Leeway config

claeyswo:

PR #616, opened 2024-11-13
feat: Feature parity for `a_decode_token` and `decode_token`

Krismix1:
I saw that the `decode_token` allows configuring the `leeway` (introduced as part of #568), but the same feature is unavailable in `a_decode_token`.
In an attempt to bridge the gap between, I refactored out the logic to validate the token, leaving the functions only responsible for fetching the public key.
Also fix the case with using `validate=False` + a `key` resulting in the validation still being executed.

There is a slight difference in the behaviour though, with regards to how `key=None` is handled:
Before:
A call like `decode_token(token, validate=True, key=None)` would be considered "okay", but fail with `jwcrypto.jws.InvalidJWSSignature: Verification failed for all signatures["Failed: [ValueError('Unrecognized key type')]"]` because the key was `None`.
With the new change:
The same call will fetch the public key and use it to validate the token.

I think it's an acceptable change in behaviour/API, as I doubt an exception was what consumers would want.

### DIFF
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -28,7 +28,7 @@ class to handle authentication and token manipulation.
 """
 
 import json
-from typing import Optional
+from typing import Optional, Union
 
 from jwcrypto import jwk, jwt
 
@@ -582,9 +582,13 @@ class KeycloakOpenID:
         return raise_error_from_response(data_raw, KeycloakPostError)
 
     @staticmethod
-    def _verify_token(token, key, **kwargs):
+    def _verify_token(token, key: Union[str, jwk.JWK, jwk.JWKSet, None], **kwargs):
         # keep the function free of IO
         # this way it can be used by `decode_token` and `a_decode_token`
+        if isinstance(key, str):
+            key = "-----BEGIN PUBLIC KEY-----\n" + key + "\n-----END PUBLIC KEY-----"
+            key = jwk.JWK.from_pem(key.encode("utf-8"))
+
         if key is not None:
             leeway = kwargs.pop("leeway", 60)
             full_jwt = jwt.JWT(jwt=token, **kwargs)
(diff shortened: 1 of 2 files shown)


src/keycloak/keycloak_openid.py
```diff
@@ -581,6 +581,25 @@ def introspect(self, token, rpt=None, token_type_hint=None):
...
 
+    @staticmethod
+    def _verify_token(token, key: Union[str, jwk.JWK, jwk.JWKSet, None], **kwargs):
+        # keep the function free of IO
+        # this way it can be used by `decode_token` and `a_decode_token`
+        if isinstance(key, str):
+            key = "-----BEGIN PUBLIC KEY-----\n" + key + "\n-----END PUBLIC KEY-----"
+            key = jwk.JWK.from_pem(key.encode("utf-8"))
+
+        if key is not None:
```
ryshoooo, line 600: I believe this is a breaking change. Previously you could specify a key and set `validate=False`, the key would have no effect and the decoded token would be returned. Now if I specify a key, validation is always done no matter what the `validate` value is. I'd suggest that the validation should only be performed if `validate=True`

Krismix1: Oh, good catch! I was too focused on the `validate=True`...

I thought the same, but actually the behavior is different on `master` - the validation will still happen. This is because `kwargs` will contain `key`, which will cause `jwcrypto` to run the validation. So `master` has a bug. This is an example fix for `master`
```python
        if validate:
            # ...
        else:
            kwargs.pop("key", None) # have to remove the key
            full_jwt = jwt.JWT(jwt=token, **kwargs)
            full_jwt.token.objects["valid"] = True
            return json.loads(full_jwt.token.payload.decode("utf-8"))
```

I can fix the bug in this PR

Krismix1: Fixed, also added some unit tests to cover it. Let me know what you think

ryshoooo: Nice catch as well! :) thanks a lot!

ryshoooo (approved):
Thanks a lot @Krismix1 LGTM

PR #621, opened 2024-11-28
fix: change httpx contract to support updated httpx to address issue 620

jonmongeau13:
.


jonmongeau13:
Will refine the change. The proxies argument is changed to `proxy` or `mounts`  (not both).

jonmongeau13:
Kept the contracts for proxies in keycloak.

Docs for httpx: https://www.python-httpx.org/api/#asyncclient
Docs for httpx proxies: https://www.python-httpx.org/advanced/proxies/

PR #622, opened 2024-11-28
fix: change httpx proxies argument to mounts

jonthemango:
Due to recent httpx library changes, `proxies` has been replaced for `proxy` and `mounts`.

Changelog for httpx: https://github.com/encode/httpx/blob/master/CHANGELOG.md#0280-28th-november-2024

Docs:
- https://www.python-httpx.org/advanced/proxies/
- https://www.python-httpx.org/api/#asyncclient

### DIFF
--- a/src/keycloak/connection.py
+++ b/src/keycloak/connection.py
@@ -102,7 +102,7 @@ class ConnectionManager(object):
         if proxies:
             self._s.proxies.update(proxies)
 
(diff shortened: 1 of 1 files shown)


allen0099:
Should we limit the httpx version to `>=0.26.0` since `mounts` only allow after this version.
https://github.com/encode/httpx/blob/master/CHANGELOG.md#0260-20th-december-2023

ryshoooo (approved):
Thanks a lot @jonthemango! :) LGTM

PR #634, opened 2025-01-27
feat: Add functions to get/update realm users profile

asyd:
Hello there, 

a small PR to add function to get/put users profiles. Needed to manage users attributes for recent Keycloak versions.

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -2297,6 +2297,24 @@ class KeycloakAdmin:
         )
         return raise_error_from_response(data_raw, KeycloakGetError, expected_codes=[HTTP_OK])
 
+    def get_realm_users_profile(self) -> dict:
+        """
+        Get list of attributes and group for given realm.
+
+        Related documentation:
+        https://www.keycloak.org/docs-api/26.0.0/rest-api/index.html#_get_adminrealmsrealmusersprofile
+
+        Return https://www.keycloak.org/docs-api/26.0.0/rest-api/index.html#UPConfig
+        :returns: UPConfig
+
+        """
+        params_path = {"realm-name": self.connection.realm_name}
+
+        data_raw = self.connection.raw_get(
+            urls_patterns.URL_ADMIN_REALM_USER_PROFILE.format(**params_path),
+        )
+        return raise_error_from_response(data_raw, KeycloakGetError, expected_codes=[HTTP_OK])
+
     def get_realm_roles(self, brief_representation: bool = True, search_text: str = "") -> list:
         """
         Get all roles for the realm or client.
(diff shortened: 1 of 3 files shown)


src/keycloak/keycloak_admin.py
```diff
@@ -2911,6 +2929,26 @@ def update_realm_role(self, role_name: str, payload: dict) -> bytes:
...
+        https://www.keycloak.org/docs-api/26.0.0/rest-api/index.html#UPConfig
+        """
+        params_path = {"realm-name": self.connection.realm_name}
+        data_raw = self.connection.raw_put(
+            urls_patterns.URL_ADMIN_REALM_USER_PROFILE.format(**params_path),
+            data=json.dumps(payload),
+        )
+        return raise_error_from_response(
+            data_raw,
+            KeycloakDeleteError,
```
ryshoooo: Shouldn't this be `KeycloakPutError`?

src/keycloak/keycloak_admin.py
```diff
@@ -5490,6 +5528,26 @@ async def a_update_realm(self, realm_name: str, payload: dict) -> dict:
...
+        https://www.keycloak.org/docs-api/26.0.0/rest-api/index.html#UPConfig
+        """
+        params_path = {"realm-name": self.connection.realm_name}
+        data_raw = await self.connection.a_raw_put(
+            urls_patterns.URL_ADMIN_REALM_USER_PROFILE.format(**params_path),
+            data=json.dumps(payload),
+        )
+        return raise_error_from_response(
+            data_raw,
+            KeycloakDeleteError,
```

tests/test_keycloak_admin.py
```diff
@@ -204,6 +204,15 @@ def test_realms(admin: KeycloakAdmin) -> None:
     assert "master" in realm_names, realm_names
     assert "test" in realm_names, realm_names
 
+    # Get and update users profile
+    user_profile = admin.get_realm_users_profile()
+    assert "attributes" in user_profile
+    assert "groups" in user_profile
+
+    res = admin.update_realm_users_profile(user_profile)
```
ryshoooo: Could we also test updating some attributes/groups? Just to be sure that we're not only hitting an edge case here as the profile is unchanged.

ryshoooo (commented):
Thanks a lot @asyd ! Overall looks good, just left a few comments :)

tests/test_keycloak_admin.py
asyd: Sure!

asyd:
I skipped test for versions < 24. I guess the API was not stable before (no easy to find KC changelog for versions around 22)

ryshoooo (approved):
Thanks a lot! LGTM

PR #637, opened 2025-02-03
Add py.typed marker file (pep-561)

PotatoesMaster:
Hi,

This change allows to use Mypy to type-check a project using python-keycloak without requiring further configuration and without raising an error
> Skipping analyzing "keycloak": module is installed, but missing library stubs or py.typed marker  [import-untyped]

Btw: thanks for your work ;)


ryshoooo (approved):
Hi @PotatoesMaster 

Thanks a lot! LGTM :)

PR #640, opened 2025-02-21
fix: add correct return type to get_credentials and a_get_credentials

Liambeck99:
Hello 👋 

Was using this earlier and noticed the return type on `get_credentials` states `-> dict` whereas its actually a list, as confirmed by the Keycloak API docs and also the Docstring stating a list, so thought I'd update 😄


ryshoooo (approved):
Hi @Liambeck99 

Thanks a lot for the fix! LGTM :)

PR #650, opened 2025-05-12
feat: adds support for group get all roles endpoint

hardenchant:
I have no tests right now, but this PR give you ability to request all bindings (clients and realm) for a group with one request, just like keycloak do it in the admin panel.


ryshoooo (commented):
Hi @hardenchant 

Thanks a lot! Looks OK, but I can't really say whether this works or not without unit tests. Could we add those in as well?

PR #651, opened 2025-05-13
feat: add pool_maxsize parameter to connection managers

### LINKED ISSUE #630: Unable to Configure Connection Pool Size in `python-keycloak`
**Context:**
I am using the `python-keycloak` library in a FastAPI-based user management application. Some of the functionalities in my application require parallel processing. During execution, I encountered the following warning:

`WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: auth.************.com. Connection pool size: 10`

After some investigation, I determined that the `KeycloakAdmin` class from the `python-keycloak` library relies on `urllib3` under the hood for HTTP requests. However, I couldn't find any documentation or examples showing how to configure the connection pool size for `urllib3` when using this library.

**Expected Behavior:**
I would like to:
1. Configure the connection pool size for `urllib3` to handle the parallel processing requirements in my application.
2. Understand if there is an existing way to adjust the connection pool size in `python-keycloak`, or if this functionality needs to be added.

Thank you for your support!

andrewmooreio:
**Description**

This PR adds a `pool_maxsize` parameter to the `KeycloakOpenID` and `KeycloakAdmin` classes. This parameter is passed to the internal `ConnectionManager`, allowing users to configure the maximum number of connections to keep in the connection pool.

**Motivation/Context**

Controlling the connection pool size is important for managing resource usage and improving performance in applications that make frequent calls to Keycloak. By exposing this parameter, users of the library can fine-tune the connection behaviour to better suit their environment and workload, potentially reducing overhead from establishing new connections.

Should resolve: https://github.com/marcospereirampj/python-keycloak/issues/630

**Testing Evidence**

- Added `pool_maxsize` parameter to relevant test fixtures in `tests/conftest.py` with a non-default value (`5`).
- Added assertions in `tests/test_keycloak_openid.py` (`test_keycloak_openid_init` and `test_a_well_known`) to verify `pool_maxsize` is correctly set on the `ConnectionManager` for `KeycloakOpenID`.
- Added assertions in `tests/test_keycloak_admin.py` (`test_keycloak_admin_init` and `test_a_realms`) to verify `pool_maxsize` is correctly set on the `ConnectionManager` for `KeycloakAdmin`.


andrewmooreio:
Note, the failed tests seem to be unrelated to this change and have been failing in CI for a while.

tests/conftest.py
ryshoooo, line 1: Can we keep the fixtures minimal and just delete the `pool_maxsize` in here? (i.e. let the defaults take place). This way we can be sure that minimal configuration will keep working for most operations.

tests/test_keycloak_admin.py
ryshoooo, line 1: Similar here. There's no need to set the pool max size for all these cases. 1 time is enough.

ryshoooo (commented):
Hi @andrewmooreio 

Thanks a lot for the changes! Overall looks good to me, I can't spot anything breaking. I've only added a comment about the tests, ideally we want to keep the fixtures as simple as possible, there is no need to set the pool size for each fixture.

ilnarkhasanov:
Hi @ryshoooo @andrewmooreio!

Can I proceed on this PR?

PR #655, opened 2025-06-05
chore: add correct return type to get_client_secrets and ...

### LINKED ISSUE #643: wrong / unspecific type annotation for some functions
I tried working with some of the api functions and seems that the output I get is different that the annotation.

a few examples,

update_client returns dict[str, Any] but annotation is bytes,

get_client_secrets returns dict[str, Any] annotation says  list

also for the cases where annotation is list (131) in keycloak_admin.py would be nice to have a complete type for the list or at least an Any, same goes for dict

let me know if I can help further

mdujava:
…a_get_client_secrets

Verified with RHBKv24 and RHBKv26

Fixes: #643

### DIFF
--- a/src/keycloak/keycloak_admin.py
+++ b/src/keycloak/keycloak_admin.py
@@ -5197,7 +5197,7 @@ class KeycloakAdmin:
         )
         return raise_error_from_response(data_raw, KeycloakPostError)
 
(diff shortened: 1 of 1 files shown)


ryshoooo (approved):
Thanks a lot @mdujava! LGTM

PR #659, opened 2025-06-22
Add pagination support to get realm roles methods

NikshetSteh:
Add pagination support to the `get_realm_roles` and `a_get_realm_roles` methods


ryshoooo (approved):
Thanks a lot @NikshetSteh! Didn't realize these were paginated :) LGTM!

PR #666, opened 2025-08-19
Prevent all HTTPX deprecation warnings

### LINKED ISSUE #646: httpx deprecation warnings
Hello.

We've been getting this warning quite a lot in our tests for some time
```
 ....\.venv\Lib\site-packages\httpx\_models.py:408: DeprecationWarning: Use 'content=<...>' to upload raw bytes/text content.
    headers, stream = encode_request(
```

I kept looking for a bad use on our code and couldn't find it and assumed it was a false report. Until today when I decided to debug it and found out that it's python-keycloak which is causing this warning. More info at https://www.python-httpx.org/compatibility/#request-content

I would be happy to make a PR unless there is a need to keep it this way, such as httpx backwards compatibility.

gregriff:
As noted in #646, people are seeing HTTPX deprecation warnings due to [this](https://www.python-httpx.org/compatibility/#request-content). In testing, this is causing hundreds of warnings. 

I isolated these warnings to the async functions, which is the only place where HTTPX is being used. Although a bit ugly, a small helper function to use the correct kwarg to `httpx.AsyncClient.request` resolved these warnings. 

Closes #646

### DIFF
--- a/src/keycloak/connection.py
+++ b/src/keycloak/connection.py
@@ -395,7 +395,7 @@ class ConnectionManager:
                 method="POST",
                 url=urljoin(self.base_url, path),
                 params=self._filter_query_params(kwargs),
-                data=data,
+                **self._prepare_httpx_request_content(data),
                 headers=self.headers,
                 timeout=self.timeout,
             )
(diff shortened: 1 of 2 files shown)


ryshoooo (approved):
Thanks a lot @gregriff! LGTM :)

PR #671, opened 2025-09-19
chore: correct return type hint in uma_permissions and a_uma_permissions

### LINKED ISSUE #663: uma_permissions wrong typehints
Type hint is Dict, but returning value is list, so mypy raise warnings when I handling response as List

```
def uma_permissions(self, token: str, permissions: str = "", **extra_payload: dict) -> dict:
```

Image

I just updated from v4 to v5,  and found this typehints

nasudadada:
## Fix uma_permissions return type hint

The `uma_permissions` method is annotated to return `dict` but actually returns `list`. This causes mypy warnings when handling the response as a list.

**Changes:**
- Update return type hint from `dict` to `list` for both sync and async methods
- Update docstring `:rtype:` annotations accordingly

**Note:**
I noticed @Ujifman reported this issue but hasn't received feedback yet. If this approach is not appropriate, please feel free to decline and provide guidance.

Fixes #663

### DIFF
--- a/src/keycloak/keycloak_openid.py
+++ b/src/keycloak/keycloak_openid.py
@@ -790,7 +790,7 @@ class KeycloakOpenID:
 
         return list(set(permissions))
 
-    def uma_permissions(self, token: str, permissions: str = "", **extra_payload: dict) -> dict:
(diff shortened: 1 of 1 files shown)


ryshoooo (approved):
Thanks a lot @nasudadada! LGTM

PR #676, opened 2025-10-16
Revert error handling change in `KeycloakAdmin.get_group_by_path`

### LINKED ISSUE #675: `KeycloakAdmin.get_group_by_path` DOES NOT raise an exception when the group was not found
#627 resolved the "issue" #617 with a solution which I believe is unacceptable. `KeycloakAdmin.get_group_by_path` no longer throws an error when the group was not found, returning the error message from the method instead. This behavior is unexpected and **undocumented**.

Calling
```python
keycloak_admin.get_group_by_path('/nonexistent_group')
```
returns
```python
{'error': 'Group path does not exist'}
```

Especially with Python being a dynamically typed language, the unexpected return value can propagate to other code and make tracing the error difficult, if the method user does not add a check for the response being containing an `error` field.

The same complaint was voiced by others in the original issue:
- https://github.com/marcospereirampj/python-keycloak/issues/617#issuecomment-2575165517
- https://github.com/marcospereirampj/python-keycloak/issues/617#issuecomment-2750924107

I believe the change from #627 should be reverted. A custom "Not found" exception can optionally be added, but the best solution would be for the method's not found error handling to match those of other methods which can fail with a 404 status code.

Please note that most fixes might affect existing code, but should not make any existing code incorrect - the same is not true for the "fix" introduced in #627.

dominik-korsa:
Fixes #675.

The reverted change was introduced in #627.

Please note that just as #627, this is a breaking change, but it might be acceptable, as raising an exception should not make existing code *incorrect*.


ryshoooo (approved):
Thanks a lot @dominik-korsa and co for bringing this back up. I believe all your points are valid, especially for consistency across the library, does not make sense for one method not to raise GET exception while others do.

Twice now we have changed the behavior of this function in a breaking manner without actually doing a major version patch. Which sucks. I do not feel like doing it for the 3rd time. I'm happy with the changes here, but I'd like to release this as a breaking change. I do understand that this is technically not breaking as it has been released already, but those releases were unforeseen with a breaking change. Had I spotted the breakage, it'd have been released as such.

Would that be agreeable?

dominik-korsa:
Sure, I also think bumping the major version with this change is the correct way to go. Thanks for the detailed response btw!

PR #696, opened 2026-01-16
chore: typehints

Ujifman:
Fixed typehint for uma permission methods, as there can be not only str


ryshoooo (approved):
Thanks a lot @Ujifman! LGTM :)

PR #697, opened 2026-01-22
chore(typehints): fixup typehints for role parameter for add composites

Ujifman:
Fixed typehint for role param for add composites endpoints.
There should be Role representations, so dict or list of dicts, not str, according to Keycloak API reference
https://www.keycloak.org/docs-api/latest/rest-api/index.html#_body_parameter_87


ryshoooo (approved):
Thanks a lot!

PR #702, opened 2026-02-11
fix: set requested resource ids correctly in permission ticket creation

### LINKED ISSUE #701: Permission ticket creation does not include resource if scope is given in permission
When creating a permission ticket with the `keycloak_uma` module, the resource_id is always None.

The resource ids are requested, but the id is only assigned to the permission but not to the variable which is used for the request. See here: https://github.com/marcospereirampj/python-keycloak/blob/ce07157bd9ef218e9d25b140c9f6d444b0c9e7b1/src/keycloak/keycloak_uma.py#L397

This results in requests like:

```javascript
{"resource_id": null, "resource_scopes": ["read", "create"]}
```

aemaem:
Fixes #701

### DIFF
--- a/src/keycloak/keycloak_uma.py
+++ b/src/keycloak/keycloak_uma.py
@@ -394,7 +394,7 @@ class KeycloakUMA:
                     msg = "Invalid resource specified"
                     raise KeycloakPostError(msg)
 
-                permission.resource_id = resource_ids[0]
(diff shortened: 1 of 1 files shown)


ryshoooo (approved):
Thanks a lot @aemaem! LGTM :)

PR #706, opened 2026-02-16
fix: correctly provide permission parameter for more than one permission in permissions_check

### LINKED ISSUE #705: "permissions_check" method uses wrong "permission" payload when permissions > 1
According to the Keycloak API the `permission` parameter must be provided multiple times when checking more than one permission (see [docs](https://www.keycloak.org/docs/26.4.7/authorization_services/index.html#_service_obtaining_permissions)).

> A string representing a set of one or more resources and scopes the client is seeking access. This parameter can be defined multiple times in order to request permission for multiple resources and scopes. This parameter is an extension to `urn:ietf:params:oauth:grant-type:uma-ticket` grant type in order to allow clients to send authorization requests without a permission ticket. The format of the string must be: `RESOURCE_ID#SCOPE_ID`. For instance: `Resource A#Scope A`, `Resource A#Scope A, Scope B, Scope C`, `Resource A`, `#Scope A`.

I had to read this documentation multiple times to understand this. So Scopes can be added comma separated but for entire permissions the parameter needs to be provided multiple times. Since the `formParams` in the [Keycloak source code](https://github.com/keycloak/keycloak/blob/38c38721ad4c1303fa2a1bc24ccece4799a8a242/services/src/main/java/org/keycloak/protocol/oidc/grants/PermissionGrantType.java#L160) is a `MultivaluedMap` only when providing the parameter multiple times, the permissions are resolved correctly for the permission decision.

The code causing the problem is here:
https://github.com/marcospereirampj/python-keycloak/blob/c5848e806d8539e5e80d01d265e230ac870572f7/src/keycloak/keycloak_uma.py#L447

aemaem:
Fixes #705

### DIFF
--- a/tests/test_keycloak_uma.py
+++ b/tests/test_keycloak_uma.py
@@ -292,8 +292,24 @@ def test_uma_access(uma: KeycloakUMA) -> None:
     assert uma.permissions_check(token["access_token"], permissions)
 
     permissions.append(UMAPermission(resource="not valid"))
+    assert uma.permissions_check(token["access_token"], permissions)
+
+    permissions = [UMAPermission(resource="not valid")]
     assert not uma.permissions_check(token["access_token"], permissions)
+
+    resource_without_a_policy = {
+        "name": "test_without_policy",
+        "scopes": ["read", "write"],
+        "type": "urn:test-no-policy",
+        "ownerManagedAccess": True,
+    }
+    resource_no_policy = uma.resource_set_create(resource_without_a_policy)
+
+    permissions = [UMAPermission(resource=resource_without_a_policy["name"])]
+    assert not uma.permissions_check(token["access_token"], permissions)
+
     uma.resource_set_delete(resource["_id"])
+    uma.resource_set_delete(resource_no_policy["_id"])
 
 
 def test_uma_permission_ticket(uma: KeycloakUMA) -> None:
--- a/tests/test_keycloak_uma.py
+++ b/tests/test_keycloak_uma.py
@@ -619,8 +619,24 @@ async def test_a_uma_access(uma: KeycloakUMA) -> None:
     assert await uma.a_permissions_check(token["access_token"], permissions)
 
     permissions.append(UMAPermission(resource="not valid"))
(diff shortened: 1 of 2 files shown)


aemaem:
Still need to fix the tests but don't know why the changes cause the current result.

tests/test_keycloak_uma.py
```diff
@@ -292,8 +292,24 @@ def test_uma_access(uma: KeycloakUMA) -> None:
     assert uma.permissions_check(token["access_token"], permissions)
 
     permissions.append(UMAPermission(resource="not valid"))
+    assert uma.permissions_check(token["access_token"], permissions)
```
aemaem, line 295: This permission check now succeeds, because the resource `not valid` is not being checked. This is because Keycloak does not include it in the check, if the resource does not exist. See the [code in Keycloak](https://github.com/keycloak/keycloak/blob/3dad921b8a19c2ddab60cd5c860f49fa3444df09/services/src/main/java/org/keycloak/authorization/authorization/AuthorizationTokenService.java#L639) to verify it: 

```java
Resource ownerResource = resourceStore.findByName(resourceServer, resourceId, identity.getId());

if (ownerResource != null) {
    permission.setResourceId(ownerResource.getId());
    addPermission(request, resourceServer, authorization, permissionsToEvaluate, limit, requestedScopesModel, ownerResource);
}
```

tests/test_keycloak_uma.py
```diff
@@ -292,8 +292,24 @@ def test_uma_access(uma: KeycloakUMA) -> None:
...
+    resource_without_a_policy = {
+        "name": "test_without_policy",
+        "scopes": ["read", "write"],
+        "type": "urn:test-no-policy",
+        "ownerManagedAccess": True,
+    }
+    resource_no_policy = uma.resource_set_create(resource_without_a_policy)
+
+    permissions = [UMAPermission(resource=resource_without_a_policy["name"])]
+    assert not uma.permissions_check(token["access_token"], permissions)
```
aemaem, line 309: Added another test here to verify, that the permission check fails, if the resource exists but is not bound to any permissive policy.

PR #707, opened 2026-02-24
Improve ConnectionManager to handle trailing slash (`/`) in `base_url` automatically 

### LINKED ISSUE #149: SUGGESTION: Force the `server_url` name to end with a '/'
I just spent a couple of hours wondering why my `KeycloakOpenID` failed to retrieve the `.well_known` endpoints.

It turned out I had forgotten to end the `server_name` (eg, `http://mykeycloak.com/auth`) by a `/`, implying that the `auth` got dropped by `urljoin` in the connector.

It could be more user-friendly to add a check at the initialization of `KeycloakOpenID`.

DanielYang59:
fix https://github.com/marcospereirampj/python-keycloak/issues/149

### DIFF
--- a/src/keycloak/connection.py
+++ b/src/keycloak/connection.py
@@ -24,11 +24,6 @@
 
 from __future__ import annotations
 
-try:
-    from urllib.parse import urljoin
-except ImportError:  # pragma: no cover
-    from urlparse import urljoin  # pyright: ignore[reportMissingImports]
-
 from typing import Any
 
 import httpx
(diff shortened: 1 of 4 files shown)


DanielYang59:
@ryshoooo can you please review this? Here I didn't normalize `base_url` in the property setter (e.g. add a trailing slash by default), but at the URL building stage, because I'm not sure if there's any case `base_url` needs to NOT have trailing slash. Thanks

Issue #711, opened 2026-03-29
Remove `deprecation` package as depedency


kurtqq:
The project has
https://github.com/marcospereirampj/python-keycloak/blob/c5848e806d8539e5e80d01d265e230ac870572f7/pyproject.toml#L37
it as dependency but I didn't find anywhere in the code where it's being used?
Regardless this package didn't have a release in the last 5 years and seems stale even if keycloak needs it there should be a replacement

kurtqq:
@NIK-TIGER-BILL I saw you handled it in another project. maybe you can help here too? thanks

PR #713, opened 2026-04-18
Remove deprecation package as depedency

kurtqq:
The deprecation package:

1. Has had no release since 2019 (last release 2.1.0 in Aug 2018)
2. Is flagged as a security / maintenance risk by multiple dependency scanners
3. Is not used in the code base

it can just be removed

closing: https://github.com/marcospereirampj/python-keycloak/issues/711


kurtqq:
@ryshoooo can you take a look?

kurtqq:
@ryshoooo can you please take a look?

PR #723, opened 2026-06-07
feat(admin): add user_logout_session to log out a single user session

astratakis:
Add user_logout_session and a_user_logout_session to KeycloakAdmin to log out a single user session, complementing user_logout which terminates all of a user's sessions. The methods validate that the session belongs to the user (for online sessions) and support removing offline sessions via is_offline.

Includes sync + async tests and a docs example; verified against Keycloak 26.0.


astratakis:
The red **latest** jobs and skipped `build` are pre-existing failures unrelated to this PR (they fail on `master` too  because the `latest` Keycloak image has drifted from the test assertions); this change passes on all pinned versions 22.0–26.0.

PR #724, opened 2026-07-29
Do not modify ConnectionManager headers for each request

dolfinus:
ConnectionManager is designed to be created as a shared object (according to name and `max_pool_size` param).

But calling `add_param_headers` / `del_param_headers` will modify headers for _all_ connections send, not just next one. There are a lot of calls like that within `KeycloakOpenID` class methods, so it is not safe to use these methods from parallel asyncio tasks.

Also if request method raised an exception, header cleanup methods are never called, potentially breaking the entire ConnectionManager.

Instead of modifying shared `headers` object, merge connection headers with per-request headers, making a local dict copy just before request is sent.

### DIFF
--- a/src/keycloak/connection.py
+++ b/src/keycloak/connection.py
@@ -252,7 +252,7 @@ class ConnectionManager:
         self._pool_maxsize = value
 
     @property
-    def headers(self) -> dict | None:
+    def headers(self) -> dict:
         """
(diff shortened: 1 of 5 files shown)


dolfinus:
`latest` Keycloak test failures are not related to this change.

PR #726, opened 2026-08-03
Migrate unmaintained `httpx` to `httpx2`

### LINKED ISSUE #725: Migrate unmaintained `httpx` to `httpx2`?
- https://github.com/openai/openai-python/issues/3375
- https://github.com/Kludex/starlette/pull/3291
> httpx has become somehow unmaintained (closing issues, discussions, and no release since 2024), and https://github.com/pydantic/httpx2 is the least annoying path forward for every consumer of that package.

DanielYang59:
close https://github.com/marcospereirampj/python-keycloak/issues/725

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -3,6 +3,7 @@
 from inspect import iscoroutinefunction, signature
 
 import pytest
+from httpx2 import AsyncClient
 
 from keycloak.connection import ConnectionManager
 from keycloak.exceptions import KeycloakConnectionError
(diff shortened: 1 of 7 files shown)


tests/test_connection.py
```diff
@@ -45,7 +47,7 @@ def test_bad_connection() -> None:
 
 
 @pytest.mark.asyncio
-async def a_test_bad_connection() -> None:
```
DanielYang59, line 48: If I understand correctly this test would not be picked up (name should be `test_xxx`), and as far as I know marking it as `async` wouldn't expand this

